PR #42, opened 2025-03-05
Performance: reducing the percentage of FFMA interleaving yields a sight performance gain, roughly 0.5%

sazczmh:
**Reducing** the percentage of FFMA interleaving yields a **sight performance gain**, roughly **0.5%**。

image

Test on H100-SXM && CUDA 12.8.


soundOfDestiny:
Great work! Thank you very much!

PR #47, opened 2025-03-07
Add kernel‘s config cache and improve TMA alignment

DXHPC:
When testing the function **deep_gemm.gemm_fp8_fp8_bf16_nt**, we observed that for certain examples, especially matrices with smaller shapes, the overhead outside of the GEMM kernel's execution is non-negligible. By using PyTorch's profiler, we found that the execution of the functions **get_col_major_tma_aligned_tensor** and **get_best_configs** introduces significant overhead.
截屏2025-03-07 11 53 19

We tried to perform memory alignment in advance  by calling  the function **get_col_major_tma_aligned_tensor** before before invoking the function **deep_gemm.gemm_fp8_fp8_bf16_nt**. But we still noticed that the alignment-related  operations were still triggered during the execution of **deep_gemm.gemm_fp8_fp8_bf16_nt**.  Additionally, the function **get_best_configs** redundantly executes for the same inputs, yielding the same results, which introduces unnecessary overhead.
First, we fixed the improper handling of 2D matrices in function **get_col_major_tma_aligned_tensor**.  Additionally, we added a cache to obtain the optimal configuration before the GEMM kernel launches, which reduces overhead for inputs of the same size.

### DIFF
--- a/deep_gemm/jit_kernels/config.py
+++ b/deep_gemm/jit_kernels/config.py
@@ -1,0 +1,99 @@
+import hashlib
+from typing import Tuple
+from .utils import get_num_sms, ceil_div, get_col_major_tma_aligned_tensor, get_m_alignment_for_contiguous_layout
+
+def is_tma_multicast_legal(n: int, block_n: int, num_tma_multicast: int, num_sms: int) -> bool:
+    if num_tma_multicast == 1:
+        return True
+    return (n % (block_n * num_tma_multicast) == 0) and num_sms % num_tma_multicast == 0
+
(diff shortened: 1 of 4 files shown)


soundOfDestiny:
Thanks!
> the overhead outside of the GEMM kernel's execution was non-negligible

Indeed. `get_col_major_tma_aligned_tensor` is for demo use and can be fused into one CUDA kernel when we want to achieve the best performance.

kzjeef:
nice job! @DXHPC

PR #54, opened 2025-03-10
Fix 2D passed in TMA scale

YLGH:
When passing in 2D input and doing unsqueeze(0), the strides don't become (aligned_m * n, 1, aligned_m), so it breaks the check for if the scale is already in TMA aligned mode.

### DIFF
--- a/deep_gemm/jit_kernels/utils.py
+++ b/deep_gemm/jit_kernels/utils.py
@@ -11,7 +11,11 @@ def set_num_sms(num_sms: int) -> None:
         num_sms: the desired maximum SM count for all GEMM kernels to use.
     """
     global _num_sms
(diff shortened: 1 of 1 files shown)


YLGH:
Also for my understanding, just curious why we dont do get_col_major_tma_aligned_tensor entirely in 2D (and flatten 3D  things to be 2D, and then expand back out)

deep_gemm/jit_kernels/utils.py
```diff
@@ -11,7 +11,11 @@ def set_num_sms(num_sms: int) -> None:
         num_sms: the desired maximum SM count for all GEMM kernels to use.
     """
     global _num_sms
-    assert 0 < num_sms <= torch.cuda.get_device_properties(device='cuda').multi_processor_count
+    assert (
```
YLGH, line 14: also what black format settings is recommended?

PR #55, opened 2025-03-11
fix cuda_graph rng check error

sleepcoo:
When using CUDA graph, deepgemm reports an error https://github.com/sgl-project/sglang/actions/runs/13778271218/job/38531702992. I found that it was triggered by CUDA's RNG check due to map_ctype. I identified the specific type and fixed the bug.


zhyncs:
Thanks chenggang!!

PR #61, opened 2025-03-13
Add kernel for fp8 gemm with 1x128 LHS scaling and 1x128 RHS scaling.

yukavio:
Add kernel for fp8 gemm with 1x128 LHS scaling and 1x128 RHS scaling for backward of weight in training.

### DIFF
--- a/deep_gemm/__init__.py
+++ b/deep_gemm/__init__.py
@@ -3,6 +3,7 @@ import torch
 from . import jit
 from .jit_kernels import (
     gemm_fp8_fp8_bf16_nt,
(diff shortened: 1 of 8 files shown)


LyricZhao:
Thanks for your contribution! This is really helpful for WGRAD cases. But I don't think merging it in current status is good: e.g. the FFMA pattern can be still found, making it as a clean/unified/maintainable impl with the original one. And we also have our internal impl waiting to push (along with some small kernels).

Later @xuzhean will follow up (about finishing in this month), and we will push our internal unified impl, so closing it right now.

PR #62, opened 2025-03-13
[Performance] Minor speed up 2% on some shapes

fzyzcjy:
Hi, thanks for the fast kernel! When checking why 4096,32768,512 shape is somehow slow, there seems to be a very tiny optimization that we may overlap store-to-smem with gemm. Thus I made a hacky prototype (as is shown in the PR). Results show that it improves ~2% on a few shapes, thus we may conditionally compile to enable them there.

I wonder whether the direction of PR looks acceptable? If so I will polish the code (cleanup messy code, extract functions, conditionally enable, etc).

Experiments: Since this is a prototype, I only did some very naive experiments. In order to make numbers more exact, each run in test_core.py is repeated by 200 times. Then, the mean and std are computed. In addition, the bandwidth delta divided by std is shown in the last column as a very vanilla statistics information.

Results (two green rows are improvements - 1.72% and 2.35% each; two small green boxes are potentially improvements or just noises or too small to be interesting):
Pasted image 20250313210040

P.S. I am playing with other very naive optimizations as well, e.g. [this](https://gist.github.com/fzyzcjy/fc0c0014a24e3c501209c38861745cf4) (will need do to optimizations later before knowing whether it speeds up or not).

### DIFF
--- a/deep_gemm/include/deep_gemm/fp8_gemm.cuh
+++ b/deep_gemm/include/deep_gemm/fp8_gemm.cuh
@@ -146,6 +146,7 @@ fp8_gemm_kernel(__nv_bfloat16* gmem_d, float* scales_b, int* grouped_layout,
 
     // Block scheduler
     uint32_t m_block_idx, n_block_idx;
+    uint32_t prev_global_idx, n_block_idx_prev;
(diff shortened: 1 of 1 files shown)


fzyzcjy:
This is my first time trying to write/modify a kernel, because I am curious about the details, and spent half a day on it. Surprisingly I found some tiny improvements and thus submitted this PR. Just now see the to-be-released optimizations in #68, and it is amazing to see experts in this field can still find out so much room to improve for such a well-optimized kernel! Thus I close this tiny PR and wait to see it.

PR #65, opened 2025-03-13
Fix get_col_major_tma_aligned_tensor to handle 2-dimensional inputs

Z-NAVY:
When the input is two-dimensional and satisfies both column-major order alignment and TMA alignment, it should be returned directly instead of undergoing further processing, as such handling would be redundant and introduce unnecessary D2D operations.

https://github.com/Z-NAVY/DeepGEMM/commit/3f92607b981fe6ae8537dc9d7b2b9972751bc280

### DIFF
--- a/deep_gemm/jit_kernels/utils.py
+++ b/deep_gemm/jit_kernels/utils.py
@@ -89,11 +89,14 @@ def get_col_major_tma_aligned_tensor(x: torch.Tensor) -> torch.Tensor:
     # NOTES: for the extreme performance, you may rewrite/fuse this function in CUDA
     assert x.dim() in (2, 3)
     remove_dim = False
+    m, n = x.shape[-2], x.shape[-1]
+    aligned_m = get_tma_aligned_size(m, x.element_size())
     if x.dim() == 2:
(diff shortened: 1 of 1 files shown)


Z-NAVY:
When input is two-dimensional, x.unsqueeze(0) adds a batch dimension but incorrectly inherits strides, causing the batch stride to reuse the original column stride instead of spanning the full 2D tensor size. This breaks the TMA-aligned stride check, leading to unnecessary transposition.

For example, the input is `x = torch.empty_strided((2,4), (1, 2), device='cuda', dtype=torch.float)`. Here, its shape is (2, 4), stride is (1, 2), and the 32-byte float elements are divisible by 16, satisfying the TMA alignment requirements. Thus, the input already meets the criteria and can be directly returned by the get_col_major_tma_aligned_tensor function. However, after processing via `x = x.unsqueeze(0)`, its shape becomes (1, 2, 4) and its stride becomes (4, 2, 4), causing the check x.stride(0) == aligned_m * n to fail. Although the subsequent processing returns a value identical to the original input, the step `aligned_x[:, :m, :] = x` introduces an unnecessary D2D copy operation.

soundOfDestiny:
Thanks!

Issue #75, opened 2025-03-25
CUDA Compilation Error with C++17 Features in test_jit.py


CasperFang:
**Description**
Running python tests/test_jit.py fails with compilation errors related to C++17 `requires` keyword and concepts. The issue was recently introduced, as tests were working correctly with commit 3b3783d06cd4d06ac4ba048633e604151d1ee535 (March 16th).

**Error Information**
Errors occur in /usr/include/c++/13/type_traits, suggesting incompatibility between CUDA compiler and C++17 features.

**Environment**
CUDA version: 12.8
GCC version: 13.3.0

soundOfDestiny:
https://github.com/deepseek-ai/DeepGEMM/pull/76

PR #76, opened 2025-03-25
Update nvcc flag c++20

YLGH:
Needed for fconcepts


YLGH:
It should fix this issue:

https://github.com/deepseek-ai/DeepGEMM/issues/75

PR #91, opened 2025-04-22
Refactor JIT compilation (+NVRTC support)

lucifer1004:
With this PR, the JIT compilation time is reduced by ~60% when using NVCC,  ~80% when using NVRTC w/o PCH, and ~90% when using NVRTC w/ PCH.

Benchmark on Xeon(R) Platinum 8480C + H100 HBM:

| m | n | k | compilation time (s) |  |  |  | TFLOPS |  |  |  |
|----|----|----|----|----|----|----|----|----|----|----|
|  |  |  | baseline | NVCC | NVRTC | NVRTC+PCH | baseline | NVCC | NVRTC | NVRTC+PCH |
| 64 | 2112 | 7168 | 5.95 | 2.12 | 0.96 | 0.35 | 154 | 154 | 153 | 153 |
| 64 | 24576 | 1536 | 6.43 | 2.55 | 1.37 | 0.78 | 230 | 230 | 231 | 230 |
| 64 | 32768 | 512 | 6.06 | 2.20 | 0.99 | 0.40 | 180 | 180 | 180 | 180 |
| 64 | 7168 | 16384 | 6.11 | 2.25 | 1.01 | 0.43 | 289 | 290 | 289 | 289 |
| 64 | 4096 | 7168 | 5.98 | 2.15 | 0.97 | 0.37 | 207 | 207 | 206 | 206 |
| 64 | 7168 | 2048 | 6.03 | 2.19 | 1.01 | 0.42 | 188 | 189 | 187 | 188 |
| 128 | 2112 | 7168 | 5.98 | 2.18 | 0.96 | 0.37 | 283 | 280 | 281 | 282 |
| 128 | 24576 | 1536 | 6.38 | 2.56 | 1.39 | 0.78 | 440 | 435 | 434 | 434 |
| 128 | 32768 | 512 | 6.00 | 2.18 | 0.99 | 0.40 | 334 | 336 | 330 | 332 |
| 128 | 7168 | 16384 | 6.24 | 2.40 | 1.18 | 0.59 | 534 | 534 | 534 | 533 |
| 128 | 4096 | 7168 | 6.04 | 2.21 | 1.03 | 0.43 | 370 | 371 | 370 | 371 |
| 128 | 7168 | 2048 | 6.19 | 2.38 | 1.19 | 0.62 | 332 | 334 | 331 | 332 |
| 4096 | 2112 | 7168 | 6.09 | 2.30 | 1.08 | 0.49 | 1053 | 1058 | 1113 | 1114 |
| 4096 | 24576 | 1536 | 6.14 | 2.33 | 1.10 | 0.50 | 1288 | 1286 | 1290 | 1289 |
| 4096 | 32768 | 512 | 6.23 | 2.41 | 1.21 | 0.61 | 913 | 912 | 912 | 911 |
| 4096 | 7168 | 16384 | 6.31 | 2.51 | 1.27 | 0.68 | 1524 | 1524 | 1459 | 1458 |
| 4096 | 4096 | 7168 | 6.28 | 2.47 | 1.27 | 0.68 | 1446 | 1445 | 1394 | 1396 |
| 4096 | 7168 | 2048 | 6.18 | 2.40 | 1.21 | 0.61 | 1240 | 1243 | 1234 | 1237 |

Note that there is some perf drop when using NVRTC due to a known bug of NVRTC which leads to extra instructions (but in the `m=4096,n=2112,k=7168` case, NVRTC version was faster, which was a bit strange). So NVCC is kept as the default compiler for now, while NVRTC can be enabled with extra env var `DG_JIT_USE_NVRTC`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Despite its lightweight design, DeepGEMM's performance matches or exceeds expert
 - [x] Shared memory swizzling for output
 - [ ] Larger block size on N (up to 256)
 - [x] MoE scheduler with TMA multicast compatibility
-- [ ] Fix TMA multicast compatibility for indivisible shapes
+- [x] Fix TMA multicast compatibility for indivisible shapes
 - [ ] Weight gradient kernels for dense models
 - [ ] Weight gradient kernels for MoE models
 - [ ] Utility kernels for MoE models (as a pre-built CUDA library)
--- a/README.md
+++ b/README.md
@@ -17,6 +17,9 @@ Despite its lightweight design, DeepGEMM's performance matches or exceeds expert
 - [ ] Larger block size on N (up to 256)
 - [x] MoE scheduler with TMA multicast compatibility
 - [x] Fix TMA multicast compatibility for indivisible shapes
+- [ ] Skip useless computation on M
+- [ ] NVRTC as a faster compiler
+- [ ] Sanitizer for testing
 - [ ] Weight gradient kernels for dense models
 - [ ] Weight gradient kernels for MoE models
 - [ ] Utility kernels for MoE models (as a pre-built CUDA library)
--- a/README.md
+++ b/README.md
@@ -12,10 +12,11 @@ Despite its lightweight design, DeepGEMM's performance matches or exceeds expert
 
 ## Roadmap
 
-- [ ] More correctness tests for grouped-contiguous layout
+- [x] More correctness tests for grouped-contiguous layout
 - [x] Shared memory swizzling for output
 - [ ] Larger block size on N (up to 256)
-- [ ] MoE scheduler with TMA multicast compatibility
+- [x] MoE scheduler with TMA multicast compatibility
+- [ ] Fix TMA multicast compatibility for indivisible shapes
 - [ ] Weight gradient kernels for dense models
 - [ ] Weight gradient kernels for MoE models
 - [ ] Utility kernels for MoE models (as a pre-built CUDA library)
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Despite its lightweight design, DeepGEMM's performance matches or exceeds expert
 
 ## News
 
-- 2025.04.09: DeepGEMM now achieves up to **1520 TFLOPS** on H800! See [#74](https://github.com/deepseek-ai/DeepGEMM/pull/74), [#78](https://github.com/deepseek-ai/DeepGEMM/pull/78), and [#81](https://github.com/deepseek-ai/DeepGEMM/pull/81) for details.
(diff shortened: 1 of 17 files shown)


lucifer1004:
With this PR, the JIT compilation time is reduced from ~10s to ~3s (NVCC, NVRTC w/o PCH), and ~500 ms when using NVRTC w/ PCH.

LyricZhao:
Thanks for this contribution, this is huge! Will take some time to review.

lucifer1004:
Some design choices open for discussion:

> How to get a kernel from a `.cubin` file?

- The most direct way is to use `cuLibraryGetKernel`, but this requires the exact symbol of the kernel, which is mangled.
- The current implementation in TensorRT-LLM (https://github.com/NVIDIA/TensorRT-LLM/blob/777c40e5fa5f08e31352a852cffb24da627e4302/cpp/include/tensorrt_llm/deep_gemm/runtime.cuh#L106-L115) uses `cuLibraryGetKernelCount`, `cuLibraryEnumerateKernels` and `cuKernelGetName` to get the FP8 GEMM kernel we need.
- A problem is that the driver APIs used in this implementation are added in CUDA 12.3. It should not be a problem since `deep_gemm` requires CUDA >= 12.3, but it is not uncommon that a user might use CUDA >= 12.3 on a machine with an out-dated driver. Most CUDA programs can still work in this case, unless newer driver APIs are used, which, unfortunately, is the case here.
- In this PR, a different method is used to avoid use of these driver APIs.
  - For `NvccCompiler`, `cuobjdump -symbols` is used to get the symbol name, which seems to be the only option.
  - For `NvrtcCompiler`, `nvrtcAddNameExpression` and `nvrtcGetLoweredName` are used to get the exact symbol of the kernel at compile time. `cuobjdump` can also be used here, but I try to avoid use of system binaries in the NVRTC path, which is meant to work in an environment w/o CUDA Toolkit.

So we have the following options:
1. Drop support for old drivers and use the driver APIs mentioned above.
2. Use `cuobjdump` for both `NvccCompiler` and `NvrtcCompiler`.
3. Use separate paths for `NvccCompiler` and `NvrtcCompiler` (current design).

lucifer1004:
It turned out that even older drivers can work with a newer CUDA container (my previous attempt was to use a newer CUDA installation from conda with an old driver), so I reverted to the original implementation of getting kernels.

yinfan98:
This is an excellent work. May I ask if there are any plans regarding when it will be merged? @LyricZhao

LyricZhao:
> This is an excellent work. May I ask if there are any plans regarding when it will be merged? @LyricZhao

I plan to merge after the 5.1 holiday.

lucifer1004:
Moved to #94

PR #112, opened 2025-06-12
Add more GPU architectures support

RayWang96:
Hi DeepGEMM Team,

This PR is submitted by NVIDIA and introduces support for more GPU architectures.

## Key Points

- MXFP8 support
  - Same 1x128 by 128x128 SF (scaling factor) recipe as original DeepGEMM, but using [UE8M0](https://docs.nvidia.com/cuda/parallel-thread-execution/#alternate-floating-point-data-formats) instead of FP32 for SF data format.
  - Same SF input layout as original DeepGEMM.
- Multi-architecture support
  - Refactored the file structure for better extensibility, working on merging other GPU architectures (e.g. Ada) into DeepGEMM.
  - Notice: code structural changes are significant, we'd like to align with you on a multi-architecture support strategy before proceeding further.

## On-going and coming soon

- Small M optimization  (~128)
- Very small M values (<64), with swap AB optimization
- Masked grouped GEMM optimization

## Long-term optimization
- Other architecture support (e.g. Ada)
- Add more input format support such as offset layout used by TRT-LLM

We are actively working on these optimizations. It aims to continuously provide support for NVIDIA's latest GPU hardware in DeepGEMM, ensuring the community can perform efficient and scalable training and inference with DeepSeek models for next-generation AI workloads.

Thanks!


LyricZhao:
Thanks for this, as we don't have SM100 GPUs, so let's make it a draft and wait for community feedback :)

zhyncs (approved):

zhyncs:
I'll help verify. Stay tuned.

CMakeLists.txt
```diff
@@ -3,8 +3,8 @@
 cmake_minimum_required(VERSION 3.10)
 project(deep_gemm LANGUAGES CXX CUDA)
 
-set(CMAKE_CXX_STANDARD 20)
-set(CMAKE_CUDA_STANDARD 20)
+set(CMAKE_CXX_STANDARD 17)
```
lucifer1004, line 6: Why change back to 17 from 20?

RayWang96: This is a conflict after rebasing. I'll resolve this later.

zhyncs:
The Blackwell part looks good to me. I believe the remaining work involves making it compatible with Hopper. Thanks for this great work!

deep_gemm/utils/layout.py
```diff
@@ -0,0 +1,170 @@
...
+    if tma_stride_check:
+        if num_groups is not None:
+            assert sf.stride(-3) == sf.stride(-1) * sf.size(-1)
+        assert sf.stride(-2) == 1
+        assert sf.stride(-1) == get_tma_aligned_size(mn, sf.element_size())
+
+    return sf
+
+
+def transform_sf_into_required_layout(sf: torch.Tensor,
```
fzyzcjy, line 136: would be great to allow skipping (INT, 128, 128) case where the data is already pre-transformed by other fused kernels. For example, my patch:

```
    should_skip_transform = (
            (sf.dtype == torch.int and gran == (1, 128) and get_device_arch() == '100a')
            or (sf.dtype == torch.int and gran == (128, 128) and get_device_arch() == '100a')
    )

    if not should_skip_transform:
        # Pre-transform checks
        check_sf_layout(sf, mn=mn, k=k, gran=gran, num_groups=num_groups)

    if should_skip_transform:
        return check_sf_layout(sf, mn=mn, k=k, gran=(1, 128), num_groups=num_groups, tma_stride_check=True,
                               type_check=torch.int)
```

fzyzcjy: update: please refer to https://github.com/sgl-project/DeepGEMM/pull/1

RayWang96: That makes sense. I'm thinking about how to skip transform in the most appropriate way, but for now, let me merge your patch as a workaround.

fzyzcjy: Yes, there may be better methods

deep_gemm/__init__.py
```diff
@@ -1,15 +1,29 @@
...
-from . import jit
-from .jit_kernels import (
-    gemm_fp8_fp8_bf16_nt,
-    m_grouped_gemm_fp8_fp8_bf16_nt_contiguous,
-    m_grouped_gemm_fp8_fp8_bf16_nt_masked,
-    wgrad_gemm_fp8_fp8_fp32_nt,
-    k_grouped_wgrad_gemm_fp8_fp8_fp32_nt,
-    ceil_div,
-    set_num_sms, get_num_sms,
-    get_col_major_tma_aligned_tensor,
```
yewentao256, line 12: `get_col_major_tma_aligned_tensor` is this function deprecated?

yewentao256: Also, if we don't import the function here, the users have to refactor the code. Eg:
```bash
from deep_gemm import calc_diff, ceil_div

->

from deep_gemm.testing.numeric import calc_diff
from deep_gemm.utils.math import ceil_div
```
Is this expected?

yewentao256: Make sense to me, thanks!

yewentao256: @RayWang96 When you said not used in SM100, does this mean it will be automatically handled in SM100 and we don't need the code like https://github.com/vllm-project/vllm/blob/dac8cc49f43f7d2639d873532a408949169821a9/vllm/model_executor/layers/quantization/fp8.py#L654?

setup.py
```diff
@@ -69,15 +81,16 @@ def prepare_includes(self):
 
     setuptools.setup(
         name='deep_gemm',
-        version='1.0.0' + revision,
```
youkaichao, line 72: for large API changes, we can bump version to 2.0?

deep_gemm/__init__.py
yewentao256: So do you mean it is recommended that we call this `get_col_major_tma_aligned_tensor` each time before we call any gemm operation?

yewentao256:
@RayWang96 Hi, will this pr be merged soon? I think my integration has been merged to the vllm main by accident https://github.com/vllm-project/vllm/pull/19820. But if we merge this pr now, I think it would be fine.

LyricZhao:
@yewentao256 Sorry, not now. We still needs 2-3 weeks to merge with the Hopper code.

yewentao256:
> @yewentao256 Sorry, not now. We still needs 2-3 weeks to merge with the Hopper code.

okk, thanks for letting me know

yewentao256:
https://github.com/vllm-project/vllm/pull/20090
Added unit test in vllm that may be helpful.

This unit test can pass on H100. But for B200 of your integration, it can not pass, so perhaps you can use this to validate the accuracy as well.

Even if we use the exact same tensor input, and just stop at the first `m_grouped_gemm_nt_contiguous` inside the `DeepGemmExperts`, the result would be different from the correct answer from H100.

RayWang96:
@yewentao256 I read and tried to debug the unit test of VLLM, and I think the problem is that the scale is not being converted to E8M0, DeepGEMM Blackwell currently supports only E8M0 scale.

The unit test is also comparing the precision of two different setups: Triton FP8 using FP32 scales vs DeepGEMM using E8M0 scales. As a result, the assert_close test fails,  I think it's expected. We should benchmark Triton FP8 and DeepGEMM FP8 against Triton BF16. In that scenario, I found their precision loss was very similar.

yewentao256:
> @yewentao256 I read and tried to debug the unit test of VLLM, and I think the problem is that the scale is not being converted to E8M0, DeepGEMM Blackwell currently supports only E8M0 scale.
> 
> The unit test is also comparing the precision of two different setups: Triton FP8 using FP32 scales vs DeepGEMM using E8M0 scales. As a result, the assert_close test fails, I think it's expected. We should benchmark Triton FP8 and DeepGEMM FP8 against Triton BF16. In that scenario, I found their precision loss was very similar.

So the correctness of DeepGemm narrow down to a smaller scope? I am thinking since this is supported on H100, to make sure the correctness for all of the models that used DeepGemm, it should still pass on B200. Actually this makes all of the unit test corresponding to DG on vllm failed

RayWang96:
> So the correctness of DeepGemm narrow down to a smaller scope? I am thinking since this is supported on H100, to make sure the correctness for all of the models that used DeepGemm, it should still pass on B200. Actually this makes all of the unit test corresponding to DG on vllm failed

@yewentao256 

First, if you directly pass an FP32 scale into DeepGEMM Blackwell, it will first be converted to an integer and then transformed into the format required by the input (see [here](https://github.com/RayWang96/DeepGEMM/blob/multi_arch_support/deep_gemm/utils/layout.py#L88)). Therefore, this is actually incorrect—the scale needs to be pre-converted to a power of two (see [here](https://github.com/RayWang96/DeepGEMM/blob/multi_arch_support/deep_gemm/utils/math.py#L33)). This might be one reason why the unit test failed. 

On this generation of GPUs, the scale must use E8M0 to fully utilize the computing power (refer to [this](https://github.com/RayWang96/DeepGEMM/blob/multi_arch_support/deep_gemm/utils/math.py#L23-L25)). In our tests, we found that this does not have much impact on the accuracy of the DeepSeek model. Therefore, using the E8M0 scale will be our most recommended approach.

Another reason why unit test failed might be that, if GEMM with FP32 scales is used as the baseline, then the results of GEMM with E8M0 scales will not align numerically. Therefore, I think the correct way to write the unit test is to compare it with the BF16 GEMM results, rather than with the FP32 scale GEMM results.

yewentao256:
@RayWang96 Thanks for your reply!
I checked it over, the main reason is that we use `torch.assert_allclose` for element-wise comparison. Just switched to your `cal_diff` function in `test_core` for a global comparison, it works now. 
Although the element-wise accuracy decreases a little bit, but the e2e accuracy works for `Qwen3`, so I think it is fine.

```
LLM_USE_DEEP_GEMM=1 lm_eval   --model vllm   --model_args "pretrained=Qwen/Qwen3-30B-A3B-FP8,max_model_len=32768,enforce_eager=True"   --trust_remote_code   --tasks gsm8k   --num_fewshot 5   --batch_size auto
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8218|±  |0.0105|
|     |       |strict-match    |     5|exact_match|↑  |0.8946|±  |0.0085|

lm_eval   --model vllm   --model_args "pretrained=Qwen/Qwen3-30B-A3B-FP8,max_model_len=32768,enforce_eager=True"   --trust_remote_code   --tasks gsm8k   --num_fewshot 5   --batch_size auto
vllm (pretrained=Qwen/Qwen3-30B-A3B-FP8,max_model_len=32768,enforce_eager=True,trust_remote_code=True), gen_kwargs: (None), limit: None, num_fewshot: 5, batch_size: auto
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.8271|±  |0.0104|
|     |       |strict-match    |     5|exact_match|↑  |0.8984|±  |0.0083|
```

tests/test_core.py
```diff
@@ -5,293 +5,100 @@
...
+        diff = calc_diff(d, ref_d)
+        assert diff < 0.001, f'{m=}, {k=}, {n=}, {major_opt=}, {diff:.5f}'
 
-            t = bench_kineto(test_func, 'fp8_gemm', suppress_kineto_output=True)
-            print(f' > Perf (m={m:5}, n={n:5}, k={k:5}): {t * 1e6:4.0f} us | '
-                  f'throughput: {2 * m * n * k / t / 1e12:4.0f} TFLOPS, '
-                  f'{(m * k + k * n + m * n * 2) / 1e9 / t:4.0f} GB/s')
+        # noinspection PyShadowingNames
+        def test_func():
+            deep_gemm.fp8_gemm_nt(a, b, d)
```
simple86, line 35: Should it be `deep_gemm.fp8_gemm_nt(a, b, d, c=c)` ?

deep_gemm/utils/math.py
```diff
@@ -0,0 +1,46 @@
...
+    return (x + y - 1) // y
+
+
+def align(x: int, y: int) -> int:
+    return ceil_div(x, y) * y
+
+
+def ceil_to_ue8m0(x: torch.Tensor):
+    assert x.view(-1).amax().item() > 0
+    return torch.pow(2.0, torch.ceil(torch.log2(x.abs())))
```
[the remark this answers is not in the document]
RayWang96: Thanks, this function is primarily intended for testing purposes. For performance optimization, a custom implementation should be considered, potentially fused into other kernels.

yewentao256:
> In our tests, we found that this does not have much impact on the accuracy of the DeepSeek model.

Hi @RayWang96 

```bash
export VLLM_ALL2ALL_BACKEND="deepep_high_throughput"
VLLM_USE_DEEP_GEMM=1 lm_eval   --model vllm   --model_args "pretrained=deepseek-ai/DeepSeek-R1,data_parallel_size=8,gpu_memory_utilization=0.95,max_model_len=16384,enable_expert_parallel=True"   --tasks gsm8k   --batch_size auto   --num_fewshot 5
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9492|±  | 0.006|
|     |       |strict-match    |     5|exact_match|↑  |0.9492|±  | 0.006|

# Nodeepgemm
|Tasks|Version|     Filter     |n-shot|  Metric   |   |Value |   |Stderr|
|-----|------:|----------------|-----:|-----------|---|-----:|---|-----:|
|gsm8k|      3|flexible-extract|     5|exact_match|↑  |0.9530|±  |0.0058|
|     |       |strict-match    |     5|exact_match|↑  |0.9522|±  |0.0059|
```

Another question is about E2E accuracy, this is the result we have on vllm, does this match your result internally?

tests/test_core.py
yewentao256: Sounds great, thanks!

RayWang96:
We've added Hopper support, it's ready to be merged into main branch, @LyricZhao.

PR #115, opened 2025-06-17
Fixed the bug in get_swizzle_mode function related to elem_size setting.

Ther-LF:
Commit Message:​​
The `get_swizzle_mode` function sets the swizzle mode for output tensors but hardcodes the element size to 2 (fp16). However, in `wgrad_gemm_fp8_fp8_fp32_nt`, the output tensor is fp32 (element size 4), causing incorrect swizzle mode selection.

​​Fix:​​
Added an is_fp32_out parameter to dynamically determine the element size (2 for fp16, 4 for fp32).

### DIFF
--- a/deep_gemm/jit_kernels/gemm.py
+++ b/deep_gemm/jit_kernels/gemm.py
@@ -17,8 +17,8 @@ def is_tma_multicast_legal(shape_dim: int, block_dim: int, num_tma_multicast: in
     return divisible and num_sms % num_tma_multicast == 0
 
 
-def get_swizzle_mode(block_n: int) -> int:
-    elem_size = 2
+def get_swizzle_mode(block_n: int, is_fp32_out: bool) -> int:
+    elem_size = 4 if is_fp32_out else 2
     for mode_bytes in (128, 64, 32):
         if (block_n * elem_size) % mode_bytes == 0:
             return mode_bytes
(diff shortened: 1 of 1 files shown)


LyricZhao:
Thanks! But actually the swizzle mode for wgrad kernels is fake, it is not used. Anyway, I will merge it for later usages.

yukuai26:
@Ther-LF 
Your changes seems cause the wgrad tests to hang. I revert them first to ensure the correctness of the kernel.
I will investigate the cause later. If you have any suggestions, feel free to let me know.

Ther-LF:
Hey @yukuai26, for the first problem size `(7168,2112)` in `test_wgrad_gemm()`, the modified `get_swizzle_mode()` returns `mode_bytes=32`. 
In https://github.com/deepseek-ai/DeepGEMM/blob/e82c4139da2e3174ff8e0e122d550e4111053369/deep_gemm/jit_kernels/wgrad_gemm.py#L82, this makes `tensor_map_d` add swizzle when generating with `make_2d_tma_d_desc()`, but `fp8_wgrad_gemm_kernel` lacks `kSwizzleDMode` param, causing kernel hang.

How about changing `tensor_map_d` in `wgrad_gemm.py` from:
`make_2d_tma_d_desc(..., smem_config[1])`
to:
`make_2d_tma_d_desc(..., 0)`?

Works for me.

PR #119, opened 2025-06-24
Add initial support for Nvidia Blackwell (SM120)

celsowm:
This change introduces the necessary compiler flags and CMake configurations to enable support for the Nvidia Blackwell SM120 architecture.

- Modified deep_gemm/jit/compiler.py to include sm_120 and compute_120 flags for NVCC and NVRTC.
- Updated CMakeLists.txt to add the new architecture flags for the build process.

Further testing on Blackwell hardware is required to validate MMA instruction compatibility and overall performance.


LyricZhao:
Thanks! But these compilation flag features are included in #112. Even if you have SM120 GEMM impls, we will also merge after #112.

PR #128, opened 2025-07-18
Update CUDA toolkits requirement

RayWang96:


JianCX:
Is there a minimum GPU driver version requirement, such as version 570 or higher? Thanks

PR #129, opened 2025-07-18
Updated submodules to use https:// vs git@

smarterclayton:
Anonymous cloning is more difficult with git@ (it requires per user
SSH keys), which makes CI automation more difficult.

I suggest reverting to using https:// as it was before #112

```
$ git clone git@github.com:fmtlib/fmt.git 
Cloning into 'fmt'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```


LyricZhao:
Thanks!

PR #135, opened 2025-07-24
Fix import-time error: "undefined symbol: _ZN3c104cuda20getCurrentCUDAStreamEa"

danthe3rd:
This patch fixes it for me

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ import shutil
 import subprocess
 from setuptools import find_packages
 from setuptools.command.build_py import build_py
-from torch.utils.cpp_extension import CppExtension, CUDA_HOME
+from torch.utils.cpp_extension import CUDAExtension, CUDA_HOME
 
 current_dir = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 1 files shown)


LJC00118:
Thanks for the fix! But we can't reproduce this bug - could you share your environment details(e.g., Torch version, CUDA version, etc.)? That’d help us understand the issue better. Appreciate it!

danthe3rd:
Hi,
I had the issue with PyTorch 2.7.1 with CUDA 12.9, as well as PyTorch 2.9 nightly (from a few weeks ago) also with CUDA 12.9.
I believe the issue might be that we don't link to `libtorch_cuda`/`c10_cuda` in the current setup.py.
The `CUDAExtension` is basically the same as `CppExtension`, but it also adds those libraries:
https://github.com/pytorch/pytorch/blob/2523e587819861cd2bf8fca738ce7875aa318482/torch/utils/cpp_extension.py#L1339-L1353
Using `CUDAExtension` could also probably simplify your setup.py because you no longer need to explicitly link with cuda/cudart

LJC00118:
Thanks!

PR #139, opened 2025-07-24
Add torch as build dependency.

yuxianq:
We import torch in the setup.py, so installing this package with pip may fail with import error.
This PR adds torch as dependency to solve this issue.


LyricZhao:
Thanks!

PR #213, opened 2025-10-12
ci: Fixes for pre-built wheels 

ko3n1g:
This brings support for NGC PyTorch containers and removes cuda11 since it doesn't seem to be compatible anymore


ko3n1g:
Passing pipeline: https://github.com/ko3n1g/DeepGEMM/actions/runs/18443482987

PR #214, opened 2025-10-12
ci: Fixes for pre-built wheels

ko3n1g:
This brings support for NGC PyTorch containers and removes cuda11 since it doesn't seem to be compatible anymore

### DIFF
--- a/.github/scripts/build.sh
+++ b/.github/scripts/build.sh
@@ -14,11 +14,7 @@ export LD_LIBRARY_PATH=/usr/local/nvidia/lib64:/usr/local/cuda/lib64:$LD_LIBRARY
 export MAX_JOBS=$([ "$MATRIX_CUDA_VERSION" == "129" ] && echo 1 || echo 2)
 export NVCC_THREADS=2
 
-if [ "$MATRIX_CUDA_VERSION" == "118" ]; then
-  export TORCH_CUDA_ARCH_LIST="7.0 7.2 7.5 8.0 8.6 8.7+PTX"
-else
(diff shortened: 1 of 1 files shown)


ko3n1g:
Here's a passing run: https://github.com/ko3n1g/DeepGEMM/actions/runs/18443482987

ko3n1g:
After merging this one, can we create and push a new tag @LyricZhao ?

LyricZhao:
> After merging this one, can we create and push a new tag @LyricZhao ?

Sure! Can you bring Python 3.8 back?

ko3n1g:
> Sure! Can you bring Python 3.8 back?

which cuda and torch version do you expect/ want to support with py3.8?

LyricZhao:
How about CUDA 12.3 + PyTorch 2.1.0?

ko3n1g:
Okay, looks like we've got it now. Here's the job that also builds a wheel for Py3.8/Cu12.3,PyT2.1.0: https://github.com/ko3n1g/DeepGEMM/actions/runs/18457600231/job/52581827327

Since support for Py3.8 is quite limited I didn't include a vast grid but just this singular configuration

ko3n1g:
I think we need to use cuda11.8 instead: https://github.com/pytorch/pytorch/blob/main/RELEASE.md#release-compatibility-matrix

ko3n1g:
@LyricZhao Finally got it working with cuda12.1: https://github.com/ko3n1g/DeepGEMM/actions/runs/18465627532/job/52608861167

LyricZhao:
LGTM, ready for merge now?

ko3n1g:
Yep!

PR #218, opened 2025-10-15
chore: Rename project to ds-deem-gemm

ko3n1g:
Unfortunately `deepgemm` also doesn't work, PyPI rejects the name since it's too close to `deep-gemm`.

So let's go with `ds_deep_gemm` (same as `pip install ds-deep-gemm`)


mprpic:
Please see also https://github.com/deepseek-ai/DeepGEMM/issues/219, these name changes introduce a lot of issues for downstream users of this library.

ko3n1g:
Closing this and will revert the package name change in a moment

PR #226, opened 2025-11-04
fix: prevent int32 overflow in k-grouped GEMM size calculations

SolenoidWGT:
### Description
Fix integer overflow in k-grouped GEMM when validating tensor sizes with large dimensions.

While training the Deepseek-v3 model using DeepGemm on an H100 machine, we encountered an error in the group GEMM kernel when using long sequences ($>256\text{k}$). Upon investigation, we found the cause was an overflow during the calculation of the product of sum_k and hidden_size.

**Root cause**: When `m` or `n` are large and multiplied with `sum_k`, the result exceeds int32 max value (2,147,483,647), causing incorrect validation or undefined behavior.

**Solution**: Cast `m` and `n` to `uint64_t` before multiplication to safely handle large matrix dimensions.

### Changes
`csrc/apis/gemm.hpp:289-290`: Cast to `uint64_t` in size assertions
  ```cpp
  // Before
  DG_HOST_ASSERT(sum_mk == m * sum_k);
  DG_HOST_ASSERT(sum_nk == n * sum_k);
  
  // After
  DG_HOST_ASSERT(sum_mk == static_cast<uint64_t>(m) * sum_k);
  DG_HOST_ASSERT(sum_nk == static_cast<uint64_t>(n) * sum_k);
  ```

### DIFF
--- a/csrc/apis/gemm.hpp
+++ b/csrc/apis/gemm.hpp
@@ -280,13 +280,13 @@ static void k_grouped_fp8_gemm_nt_contiguous(const std::pair<torch::Tensor, torc
 
     // Shape checks
     const auto& [num_groups, m, n] = get_shape<3>(d);
-    const auto& sum_mk = a.first.numel();
-    const auto& sum_nk = b.first.numel();
-    int sum_k = 0;
+    const auto& sum_mk = static_cast<uint64_t>(a.first.numel());
+    const auto& sum_nk = static_cast<uint64_t>(b.first.numel());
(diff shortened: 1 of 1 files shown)


zheanxu:
Thank you for submitting this PR! We found that the same int overflow issue with `sum_k * m` also occurs in `sm90_fp8_gemm_1d1d.cuh`. This bug has been fixed in https://github.com/deepseek-ai/DeepGEMM/commit/bb4424aad49927283349738dfa7b54f5baad6025.

PR #284, opened 2026-02-02
Windows support

tori29umai0123:
## Summary

  This PR adds Windows/MSVC build support for DeepGEMM while maintaining full compatibility with existing Linux/GCC
  builds.

  ## Changes

  ### MSVC Compatibility Layer
  - Add `csrc/utils/msvc_compat.hpp` with POSIX compatibility macros (`popen`, `pclose`, `WEXITSTATUS`)
  - Add `DG_MSVC` preprocessor macro for platform-conditional compilation

  ### Code Fixes for MSVC
  - Convert C++20 designated initializers to positional brace initialization (MSVC requires aggregate types)
  - Fix `std::filesystem::path` handling (use `.string()` for conversions, as `c_str()` returns `wchar_t*` on Windows)
  - Add `.exe` suffix for Windows executables (nvcc.exe, cuobjdump.exe)

  ### Build System
  - Update `setup.py` with Windows-specific compiler flags (`/std:c++latest`, `/O2`, `/EHsc`, etc.)
  - Add cuBLASLt and cuBLAS library linking
  - Fix wheel platform tag (lowercase `win_amd64`)
  - Fix wheel filename format to comply with PEP 427

  ### Documentation
  - Add Windows build instructions to README.md

  ## Testing

  - Successfully built on Windows 11 with:
    - Visual Studio 2022 (MSVC 14.43)
    - CUDA Toolkit 12.8
    - PyTorch 2.10
    - Python 3.10

  ## Notes

  - All changes use conditional compilation to preserve Linux/GCC compatibility
  - No changes to kernel code or CUDA device code


rogeroberg:
I noticed that the same Windows path-type issue also appears in a few JIT loader call sites that may be worth including in this PR.

In particular, `std::filesystem::path::c_str()` returns `path::value_type*`, which is `wchar_t*` on Windows, while these call sites expect narrow `char*` strings:

`csrc/jit/handle.hpp`

* `cudaLibraryLoadFromFile(..., cubin_path.c_str(), ...)`
* `lazy_cuLibraryLoadFromFile(..., cubin_path.c_str(), ...)`
* `printf(..., dir_path.c_str(), ...)`
* `lazy_cuModuleLoad(..., cubin_path.c_str())`

`csrc/jit/kernel_runtime.hpp`

* several `printf("%s", path.c_str())` calls
* `fmt::format(..., cuobjdump_path.c_str(), cubin_path.c_str())`

Using `.string()` at these narrow-string call sites would avoid the `wchar_t*` / `char*` mismatch on Windows.

Not sure whether these are already covered by the current PR diff, but I wanted to flag them in case they were missed.

PR #288, opened 2026-02-22
fix: No module named 'scripts.generate_pyi'

HollowMan6:
When run `install.sh` directly, it will fail with
the following error:

```bash
  File "DeepGEMM/setup.py", line 19, in <module>
    from scripts.generate_pyi import generate_pyi_file
ModuleNotFoundError: No module named 'scripts.generate_pyi'
```

This is because the `scripts/` directory has no `__init__.py`, so Python can't import it as a package. The fix is to create one.


kibitzing:
This works for me. Thank you!

It worked properly after I committed the changes.

ajwise9:
That doesn't look like a pull request. Please re-open it in the issues section than the pull requests.

HollowMan6:
> That doesn't look like a pull request. Please re-open it in the issues section than the pull requests.

I'm not sure if I understand you fully but having an empty `scripts/__init__.py` does solve the issue I mentioned in this PR description, so this is a PR.

PR #303, opened 2026-04-11
Eliminate redundant global stores in SM90 MQA logits kernels

Gregory-Pereira:
Seeks to address: #279


llag9810:
Hi, I noticed this PR was closed. Could you share why? I am seeing a similar slowdown with fp8_mqa_logits on long-context prefill and want to understand the reason. Thanks.

PR #311, opened 2026-04-19
fix: correct fused KV cache stride in paged MQA logits

JasonOA888:
## Bug

`fp8_fp4_paged_mqa_logits` decomposes the fused KV cache into separate `kv_cache` and `kv_cache_sf` tensors via `torch::from_blob`. The strides passed to `from_blob` assume a **flat layout** (all values then all SF), but the API assertion enforces an **interleaved layout** (`[values | SF]` per row):

```cpp
DG_HOST_ASSERT(fused_kv_cache.stride(1) == fp4_with_sf_bytes);  // FP4: 68
DG_HOST_ASSERT(fused_kv_cache.stride(1) == head_dim_with_sf);  // FP8: 132
```

### What was wrong

**kv_cache stride(1):**
- FP4: was `head_dim/2` (64), should be `fp4_with_sf_bytes` (68)
- FP8: was `head_dim` (128), should be `head_dim_with_sf` (132)

**kv_cache_sf offset:**
- Was `block_kv * head_dim`, pointing past all data rows into undefined territory
- Should be `head_dim` (or `head_dim/2` for FP4), pointing to the SF of the first row

**kv_cache_sf stride(1):**
- Was `1` (contiguous), but SF values are interleaved every `fp4_with_sf_bytes`/`head_dim_with_sf` bytes

### Impact

TMA descriptors constructed with wrong strides cause misaligned reads for any row > 0 in the paged KV cache. This silently corrupts attention logits for multi-row blocks.

The test helper had the same flat-layout assumption, masking the bug for small block sizes where corruption fell in the last 1-2 rows.

## Fix

- Use the correct row stride (`fp4_with_sf_bytes` / `head_dim_with_sf`) for both `kv_cache` and `kv_cache_sf` from_blob
- Fix `kv_cache_sf` offset to `head_dim` / `head_dim / 2` (start of first row's SF)
- Fix test helpers to construct proper interleaved layout

### DIFF
--- a/csrc/apis/attention.hpp
+++ b/csrc/apis/attention.hpp
@@ -263,13 +263,13 @@ static torch::Tensor fp8_fp4_paged_mqa_logits(const std::tuple<torch::Tensor, st
         kv_cache = torch::from_blob(
             fused_kv_cache.data_ptr(),
             {num_kv_blocks, block_kv, head_dim / 2},
-            {kv_cache_stride_bytes, head_dim / 2, 1},
+            {kv_cache_stride_bytes, fp4_with_sf_bytes, 1},
             torch::TensorOptions().dtype(kPackedFP4)
         );
         kv_cache_sf = torch::from_blob(
-            fused_kv_cache.data_ptr<uint8_t>() + block_kv * head_dim / 2,
+            fused_kv_cache.data_ptr<uint8_t>() + head_dim / 2,
             {num_kv_blocks, block_kv},
-            {kv_cache_stride_bytes / static_cast<int>(sizeof(int)), 1},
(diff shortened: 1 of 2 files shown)


zheanxu:
Thanks for the detailed analysis!

However, this is actually intentional. The layout is **flat**, not interleaved — all fp8/fp4 data comes first, followed by all scale factors. The fused tensor is shaped `[block_size, head_dim_with_sf]` so that `size(0)` directly reflects the real block size, while keeping both portions in a single allocation.

The `from_blob` calls correctly split this flat buffer into two views. The proposed changes would break this by assuming an interleaved layout.

PR #324, opened 2026-05-01
feat: add sm120 support for DeepGEMM 

leavelet:
This PR adds first-class `sm_120` support to DeepGEMM, maintained on the `nv_dev` branch by the NVIDIA DevTech APAC team. It brings the full DeepGEMM kernel surface — dense, grouped/MoE, einsum, hyper-connection, and MQA-logits — to the sm120 and sm121 devices like RTX Pro 6000 and DGX Spark.

co-authored with @lucifer1004.

## Feature surface

**Dense GEMM** (`sm120_fp8_fp4_gemm_1d1d`, `sm120_bf16_gemm`)
- FP8 (e4m3, UE8M0 block-scaled, 1D1D), all layouts NT/NN/TN/TT
- FP4 (e2m1, UE8M0 `mxf4nvf4`, gran_k=32)
- BF16 and TF32
- Mixed precision: FP8_A × FP4_B (`mxf8f6f4 .e4m3.e2m1`) and FP4_A × FP8_B
  (`.e2m1.e4m3`) — both directions
- AB-swap small-M path (M = 1..16) via TMA + per-element strided epilogue
- Wave-packing latency heuristic (BLOCK_M 64/128) and split-K for small dims

**Grouped GEMM / MoE**
- M-grouped contiguous, M-grouped masked, K-grouped contiguous
- FP8 / FP4 / BF16

**Batched GEMM / Einsum** (`sm120_bmk_bnk_mn`)
- K-major B (`bhr,hdr->bhd`) and MN-major B (`bhd,hdr->bhr`)
- Split-S reduction (`bmk,bnk->mn`); FP8 / BF16; small-M AB-swap

**HC Prenorm** (`sm120_tf32_hc_prenorm_gemm`)
- Fused TF32 GEMM + square-sum with split-K

**MQA Logits ** (`sm120_fp8_mqa_logits`, `sm120_fp4_mqa_logits` + paged)
- FP8 / FP4; dense (ragged, warp-specialized, L2-cached KV, split-KV) and
  paged (next_n / kPadOddN support)

## Performance (RTX PRO 6000 Blackwell, single GPU, `tests/test_*.py`)

Roofline used: FP8 ≈ 814 TFLOPS, FP4 ≈ 1628 TFLOPS (2× FP8), BF16 ≈ 445 TFLOPS. GB/s is effective bandwidth (cache-inclusive for small/cached shapes).

### Dense GEMM
| dtype | peak TFLOPS | % roofline | peak GB/s |
|-------|------------:|-----------:|----------:|
| FP8   | 778         | 96%        | 1235      |
| FP4   | 1561        | 96%        | 1337      |
| BF16  | 444         | ~98%       | 1497      |

### Grouped GEMM (MoE)
| variant                       | FP8 TFLOPS | FP4 TFLOPS | peak GB/s |
|-------------------------------|-----------:|-----------:|----------:|
| M-grouped contiguous (prefill)| 814        | 1571       | —         |
| M-grouped masked              | 670        | 1363       | 836       |
| K-grouped contiguous (EP)     | 543 (gk32) | 666 (gk128)| 1011      |

### Einsum
Peak 742 TFLOPS (`b=8192,h=8,r=4096,d=1024`); up to 2893 GB/s (small-batch, cache-bound); Split-S path scales down for small problems.

### HC Prenorm
`m=8192, n=24, k=28672`, splits=16: 29 TFLOPS / 1248 GB/s (bandwidth-bound).

### MQA Logits
| variant     | peak TFLOPS | 
|-------------|------------:|
| Ragged FP4  | 1049        | 
| Paged FP4   | 685         |
| Ragged FP8  | 636        | 
| Paged FP8   | 547        |


linjiapro:
@leavelet  this is nice, after it is merged into the nv-dev branch, should https://github.com/vllm-project/vllm/pull/41834 merge too in order for vllm to be able to work with the branch.

jasl:
I add my benchmark at https://github.com/vllm-project/vllm/pull/41834 as references

Rachmanino:
nice work! may I ask the hardware for testing here is either 5090 or RTX6000pro?

leavelet:
@lucifer1004 @RayWang96 Please review, thanks!

brandonmmusic-max:
@leavelet — testing your SM120 `paged_mqa_logits` on **sm_120a (RTX PRO 6000)** with DeepSeek-V4 MTP `num_speculative_tokens=2` (→ `next_n=3`), I hit a CUDA illegal memory access in the kernel. Root-caused it to an **incomplete odd-`next_n` (`kPadOddN`) port**: the odd leftover atom (last atom of each next_n group, holding **1** real token, not `kNextNAtom=2`) is mishandled in three coordinated places, plus a loop bound and the host gate. All dormant at `next_n ∈ {1,2}`, firing at odd `next_n ≥ 3`:

**1. `kNumNextNAtoms` has a spurious `+ 1`** — `sm120_fp8_paged_mqa_logits.cuh:151` (FP4 `:157`):
```cpp
static constexpr uint32_t kNumNextNAtoms =
    kPadOddN ? (kNextN + kNextNAtom - 1) / kNextNAtom + 1 : (...);   // = 3 for next_n=3
```
The metadata kernel builds the schedule with stride `ceil_div(next_n, 2) = 2`, so the consumer decodes that stride-2 schedule on the wrong stride → wrong `block_table`/`context_lens` indices. SM100 uses `math::constexpr_ceil_div(kNextN, kNextNAtom)` with no `+1`.

**2. `issue_tma_q` loads 2 tokens for the odd atom — this is the actual sanitizer error.** Every atom unconditionally `tma::copy<kHeadDim, kNextNAtom * kNumHeads, ...>` (2 tokens of Q), but the odd atom only has 1 valid token, so the kernel's later compute reads the never-populated 2nd-token SMEM slot:
```
========= Invalid __shared__ read of size 16 bytes
=========     at void deep_gemm::sm120_fp8_paged_mqa_logits<(unsigned int)3, ...>
```
Fix: for the odd atom, zero the padding SMEM and TMA-copy only `kNumHeads` (1 token) of Q / weights / (FP4) scale-factors via dedicated one-token descriptors `tensor_map_q_odd` / `tensor_map_weights_odd` / `tensor_map_sf_q_odd` (built host-side in `smxx_fp8_fp4_paged_mqa_logits.hpp`).

**3. The non-varlen store dispatch is missing the `kPadOddN` arm** — `:336-343` (FP4 `:347-354`):
```cpp
} else {
    compute_and_store(cute::Int<kNextNAtom>{});   // unconditional 2 tokens, no kPadOddN arm
}
```
so the odd leftover atom stores a 2nd row at logits row `B*next_n`, one past the `[B*next_n, max_model_len]` tensor → OOB write. (SM100 has the `else if constexpr (kPadOddN) { ... Int<1> ... }` arm.)

Plus the **`q_idx` previous-q sentinel** is initialized to `batch_size` but must count atoms → `batch_size * kNumNextNAtoms` (2 sites per kernel), and the **host gate** in `csrc/apis/attention.hpp` asserts `next_n <= 2` on `arch_major == 12` (needs widening to `<= 3`). The FP4 twin carries the identical pattern throughout.

The tell: `kPadOddN` is defined in the SM120 kernel but the odd-atom load/store handling that should hang off it was never finished — the count half made it in (with the stray `+1`), but the one-token load and the `Int<1>` store didn't.

**Fix summary** (complete the odd-atom path, both FP8 + FP4): (a) drop the `+1` in `kNumNextNAtoms`; (b) add a `kPadOddN` block in `issue_tma_q` that loads one token via the `_odd` descriptors + zeroes the pad; (c) add the `kPadOddN` `Int<1>` store arm; (d) init `q_idx` to `batch_size * kNumNextNAtoms`; (e) widen the host gate + build the `_odd` descriptors host-side. Using `q_idx`, the SM120 math-warp atom index, where SM100 uses `q_atom_idx`.

One honest caveat: the `+1` and the missing store arm match a static SM100-vs-SM120 diff, but for the odd-atom **load** I went with new one-token `_odd` descriptors as a working fix for the SMEM-read OOB — I haven't verified SM100 loads the odd atom by this exact mechanism, so you may prefer a different approach there (e.g. an OOB-fill flag on the main descriptor).

**Validated on the clean `leavelet/DeepGEMM:sm120 @ 1845516…` checkout (sm_120a, RTX PRO 6000):**
- `compute-sanitizer --tool memcheck`, `next_n=3`: **before** = `Invalid __shared__ read of 16 bytes` in `sm120_fp8_paged_mqa_logits`, **34 errors** → **after** = **`ERROR SUMMARY: 0 errors`** (FP8 **and** FP4).
- `tests/test_attention.py::test_paged_mqa_logits` with `next_n=3` added to the SM120 arm (`(1,2) → (1,2,3)`): all `VALIDATION_OK` (FP8 + FP4).
- End-to-end DeepSeek-V4-Flash native `next_n=3` (on a downstream unholy-fusion build, kernel differs slightly): estonia 30-shot retrieval **30/30**; ~185 tok/s, no regression.

Reproduce + writeup + the validated 5-file diff (`dg_clean_nextn3_fix.diff`): **https://github.com/brandonmmusic-max/deepseek-sm120**. Happy to open this as a PR against `leavelet:sm120` (the kernel + host fix + the `next_n=3` test) if you'd like it folded into #324 — your call whether to take it as a commit or have me PR it. (Context: this is the sm_120a counterpart of the SM90 `next_n=3` work in vLLM #43457; SM90 uses a different, non-atom design, so this defect is SM120-specific.)

leavelet:
> @leavelet — testing your SM120 `paged_mqa_logits` on **sm_120a (RTX PRO 6000)** with DeepSeek-V4 MTP `num_speculative_tokens=2` (→ `next_n=3`), I hit a CUDA illegal memory access in the kernel. Root-caused it to an **incomplete odd-`next_n` (`kPadOddN`) port**: the odd leftover atom (last atom of each next_n group, holding **1** real token, not `kNextNAtom=2`) is mishandled in three coordinated places, plus a loop bound and the host gate. All dormant at `next_n ∈ {1,2}`, firing at odd `next_n ≥ 3`:
> 
> **1. `kNumNextNAtoms` has a spurious `+ 1`** — `sm120_fp8_paged_mqa_logits.cuh:151` (FP4 `:157`):
> 
> ```c++
> static constexpr uint32_t kNumNextNAtoms =
>     kPadOddN ? (kNextN + kNextNAtom - 1) / kNextNAtom + 1 : (...);   // = 3 for next_n=3
> ```
> 
> The metadata kernel builds the schedule with stride `ceil_div(next_n, 2) = 2`, so the consumer decodes that stride-2 schedule on the wrong stride → wrong `block_table`/`context_lens` indices. SM100 uses `math::constexpr_ceil_div(kNextN, kNextNAtom)` with no `+1`.
> 
> **2. `issue_tma_q` loads 2 tokens for the odd atom — this is the actual sanitizer error.** Every atom unconditionally `tma::copy<kHeadDim, kNextNAtom * kNumHeads, ...>` (2 tokens of Q), but the odd atom only has 1 valid token, so the kernel's later compute reads the never-populated 2nd-token SMEM slot:
> 
> ```
> ========= Invalid __shared__ read of size 16 bytes
> =========     at void deep_gemm::sm120_fp8_paged_mqa_logits<(unsigned int)3, ...>
> ```
> 
> Fix: for the odd atom, zero the padding SMEM and TMA-copy only `kNumHeads` (1 token) of Q / weights / (FP4) scale-factors via dedicated one-token descriptors `tensor_map_q_odd` / `tensor_map_weights_odd` / `tensor_map_sf_q_odd` (built host-side in `smxx_fp8_fp4_paged_mqa_logits.hpp`).
> 
> **3. The non-varlen store dispatch is missing the `kPadOddN` arm** — `:336-343` (FP4 `:347-354`):
> 
> ```c++
> } else {
>     compute_and_store(cute::Int<kNextNAtom>{});   // unconditional 2 tokens, no kPadOddN arm
> }
> ```
> 
> so the odd leftover atom stores a 2nd row at logits row `B*next_n`, one past the `[B*next_n, max_model_len]` tensor → OOB write. (SM100 has the `else if constexpr (kPadOddN) { ... Int<1> ... }` arm.)
> 
> Plus the **`q_idx` previous-q sentinel** is initialized to `batch_size` but must count atoms → `batch_size * kNumNextNAtoms` (2 sites per kernel), and the **host gate** in `csrc/apis/attention.hpp` asserts `next_n <= 2` on `arch_major == 12` (needs widening to `<= 3`). The FP4 twin carries the identical pattern throughout.
> 
> The tell: `kPadOddN` is defined in the SM120 kernel but the odd-atom load/store handling that should hang off it was never finished — the count half made it in (with the stray `+1`), but the one-token load and the `Int<1>` store didn't.
> 
> **Fix summary** (complete the odd-atom path, both FP8 + FP4): (a) drop the `+1` in `kNumNextNAtoms`; (b) add a `kPadOddN` block in `issue_tma_q` that loads one token via the `_odd` descriptors + zeroes the pad; (c) add the `kPadOddN` `Int<1>` store arm; (d) init `q_idx` to `batch_size * kNumNextNAtoms`; (e) widen the host gate + build the `_odd` descriptors host-side. Using `q_idx`, the SM120 math-warp atom index, where SM100 uses `q_atom_idx`.
> 
> One honest caveat: the `+1` and the missing store arm match a static SM100-vs-SM120 diff, but for the odd-atom **load** I went with new one-token `_odd` descriptors as a working fix for the SMEM-read OOB — I haven't verified SM100 loads the odd atom by this exact mechanism, so you may prefer a different approach there (e.g. an OOB-fill flag on the main descriptor).
> 
> **Validated on the clean `leavelet/DeepGEMM:sm120 @ 1845516…` checkout (sm_120a, RTX PRO 6000):**
> 
> * `compute-sanitizer --tool memcheck`, `next_n=3`: **before** = `Invalid __shared__ read of 16 bytes` in `sm120_fp8_paged_mqa_logits`, **34 errors** → **after** = **`ERROR SUMMARY: 0 errors`** (FP8 **and** FP4).
> * `tests/test_attention.py::test_paged_mqa_logits` with `next_n=3` added to the SM120 arm (`(1,2) → (1,2,3)`): all `VALIDATION_OK` (FP8 + FP4).
> * End-to-end DeepSeek-V4-Flash native `next_n=3` (on a downstream unholy-fusion build, kernel differs slightly): estonia 30-shot retrieval **30/30**; ~185 tok/s, no regression.
> 
> Reproduce + writeup + the validated 5-file diff (`dg_clean_nextn3_fix.diff`): **https://github.com/brandonmmusic-max/deepseek-sm120**. Happy to open this as a PR against `leavelet:sm120` (the kernel + host fix + the `next_n=3` test) if you'd like it folded into #324 — your call whether to take it as a commit or have me PR it. (Context: this is the sm_120a counterpart of the SM90 `next_n=3` work in vLLM #43457; SM90 uses a different, non-atom design, so this defect is SM120-specific.)

I'll fix this issue in this PR, thanks a lot!

leavelet:
@brandonmmusic-max Thanks for your report!

I've committed a fix (76e93aa), which will solve the problem elegantly, by aligning the SM120 API with SM100. But I did not add the one-token _odd descriptors. I kept the unconditional kNextNAtom-token TMA Q load, exactly like SM100. The last batch's odd 2nd token is out-of-bounds and zero-filled by TMA (CU_TENSOR_MAP_FLOAT_OOB_FILL_NONE), and once the store arm processes only token 0, that 2nd Q slot is loaded but never read — so the dedicated descriptors aren't needed.

My read is that the actual trigger of the Invalid `__shared__` read of 16 bytes was the q_idx sentinel (your fix d): at kNumNextNAtoms >= 2 the old batch_size sentinel collides with a valid atom index, so the math warp's first task skips the Q-pipeline wait() and reads smem_q with an uninitialized q_stage_idx. Your _odd descriptors were a sound belt-and-suspenders fix, but the read goes away once (d) + the store arm land, which is why I matched SM100 here and kept the descriptor surface minimal.

One step further on the host gate: instead of widening to next_n <= 3, I removed the SM120 cap entirely. SMEM is sized per-atom (≤2 tokens) and independent of next_n, so there's no SM120-specific reason to cap it — SM120 is now a full peer of SM100. The test is unified: arch_major in (10, 12) shares the (1, 2, 3, 4, 5, 6) next_n arm.

Thanks again — please feel free to put it through its paces, empower more SM120 devices, and send over any further feedback!

brandonmmusic-max:
Your more generalized fix is the right call.   You certainly have the holistic POV.   If I’m trying to be helpful I. Debugging I try to be targeted instead if taking a flamethrower to other people’s work lol because there are probably structural assumptions in making that I don’t even realize.  Like the smem sized per atom I wasn’t thinking from that perspective (and it was also late at night for me lol).   I’ll put this through it paced though thanks for responding!

leavelet:
@RayWang96 I have cleaned up the code and verified them on sm90, sm100 and sm120, the PR is ready for merge. Thanks!

leavelet:
@lucifer1004 @AliceChenyy cc @samuellees 

Please switch to commit 33a715e for better MoE EP performance, especially for decode. The commit bump alone gives a 2x kernel speedup, and swapAB is an optional add-on worth roughly 6%

csrc/apis/gemm.hpp
```diff
@@ -79,24 +81,95 @@ static void fp8_fp4_gemm_nt(const std::pair<torch::Tensor, torch::Tensor>& a,
     if (early_return(m, n, k, d, c))
         return;
 
-    // Transform SFA and SFB into compute-required layout
-    const auto [sfa, sfb, gran_k_a, gran_k_b] = layout::transform_sf_pair_into_required_layout(
-        a.second, b.second, m, n, k, recipe, recipe_a, recipe_b, std::nullopt, std::nullopt, disable_ue8m0_cast);
+    // Dispatch into different arch implementations
+    if (arch_major == 12) {
```
RayWang96: The SM120 branch in fp8_fp4_gemm_nt looks a bit odd to me — it's a big inline block, and it's placed before SM90/SM100 (so the order ends up 12 → 9/10). Maybe pull it out into a fp8_fp4_gemm_nt_sm120(...) helper? And keep the top level compact and in natural arch order.

tests/test_einsum.py
```diff
@@ -46,7 +46,7 @@ def test_bhr_hdr_bhd():
             ref_z = torch.einsum('bhr,hdr->bhd', x, y)
             z = torch.empty((b, h, d), device='cuda', dtype=torch.bfloat16)
             deep_gemm.einsum('bhr,hdr->bhd', x, y, z)
-            assert calc_diff(z, ref_z) < 1e-10
+            assert calc_diff(z, ref_z) < 1e-7
```
RayWang96: What's the justification for the relaxation? Is it driven by SM120 not passing at 1e-10? If so, what makes SM120's einsum less accurate here.

csrc/jit_kernels/heuristics/runtime.hpp
```diff
@@ -44,14 +44,21 @@ class HeuristicsRuntime {
...
+                                                                    const std::optional<int>& num_groups = std::nullopt) {
+        const auto arch_major = device_runtime->get_arch_major();
+        if (arch_major != 10 and arch_major != 12)
             return kLegacyMKAlignmentForContiguousLayout;
 
-        int block_m = 240, mma_step = 16;
+        // SM120: kMWarps=4, MMA_M=16, valid BLOCK_M must be multiple of 64
+        int block_m = arch_major == 12 ? 128 : 240;
+        int min_block_m = arch_major == 12 ? 64 : 32;
+        int mma_step = arch_major == 12 ? 64 : 16;
```
RayWang96: Maybe we can extract a standalone helper to help SM120 get the config.

tests/test_einsum.py
leavelet: The test compares DeepGEMM's bf16 output against cuBLAS's; both use identical HMMA.16816.F32.BF16 + a single bf16 cast, so they differ only in FP32 K-reduction order. At small M, cuBLAS's nvjet kernel sums 3 partial accumulators (vs DeepGEMM's single linear accumulator), which is marginally more accurate and makes ~0.2% of low-magnitude outputs differ, exceeding the near-bit-exact 1e-10 (lands ~`8e-9). Against an FP64 reference DeepGEMM and cuBLAS are equally accurate (~1.4e-6, ratio 1.00). I will propose a fix, keeps 1e-10for SM90/SM100, loosens to1e-7 only on SM120, and adds an FP32-reference correctness check there (< 1e-5`). Does this sound correct?

ormandj:
A heads-up that this PR has become a prerequisite for SM120, not just a nice-to-have. Now that [vllm-project/vllm#43477](https://github.com/vllm-project/vllm/pull/43477) has merged into vLLM main (2026-06-22), vLLM main routes the SM120 FP8 path through DeepGEMM. The DeepGEMM revision vLLM currently pins doesn't have SM120 (`arch_major == 12`) support yet, so the two are out of sync — and starting a stock FP8 model on an RTX PRO 6000 crashes during startup warmup until this support lands.

We ran into it bringing up `Qwen3.6-27B-FP8`. Building vLLM main at `6af0559d` (with [#43477](https://github.com/vllm-project/vllm/pull/43477) in):

```
  File ".../vllm/model_executor/warmup/kernel_warmup.py", line 67, in kernel_warmup
    deep_gemm_warmup(model, max_tokens)
  File ".../vllm/model_executor/warmup/deep_gemm_warmup.py", line 406, in deep_gemm_warmup
    deepgemm_fp8_gemm_nt_warmup(model, max_tokens, pbar)
  File ".../vllm/model_executor/warmup/deep_gemm_warmup.py", line 227, in _deepgemm_fp8_gemm_nt_warmup
    fp8_gemm_nt(
  File ".../vllm/utils/deep_gemm.py", line 453, in fp8_gemm_nt
    return _fp8_gemm_nt_impl(*args, disable_ue8m0_cast=not use_ue8m0, **kwargs)
RuntimeError: Assertion error (/workspace/.deps/deepgemm-src/csrc/apis/../jit_kernels/impls/../heuristics/../../utils/layout.hpp:76): Unknown recipe
```

What's going on: vLLM's default build fetches DeepGEMM via FetchContent, and `deepgemm.cmake` still pins `891d57b4`. There, `get_default_recipe()` only handles `arch_major` 9 and 10 and otherwise falls through to `DG_HOST_UNREACHABLE("Unknown recipe")` — which is the assertion above on `arch_major == 12`. This PR adds the `arch_major == 12` branch, which resolves it.

One thing worth flagging on scope: it isn't DSv4/GLM-specific, but it also isn't *all* FP8. vLLM only routes a linear through DeepGEMM for **block-scale FP8** — i.e. checkpoints quantized with a `weight_block_size` (the 128-block DeepSeek-style scales), where `_fp8_linear_may_use_deep_gemm()` sees `block_quant=True`. Per-tensor and per-channel FP8 take the cutlass/cublas path and aren't affected. So this hits any *block-scale* FP8 model on SM120 regardless of architecture (we saw it with `Qwen3.6-27B-FP8`, TP=1, `--kv-cache-dtype fp8`). Until this lands — or vLLM is pointed at a supporting tree via `DEEPGEMM_SRC_DIR` — block-scale FP8 inference on SM120 with current vLLM main is broadly broken, not just for DSv4/GLM.

Thanks a lot for the work here — happy to put it through its paces once it's in.

Env: vLLM `v0.23.1rc1.dev358+g6af0559dd`, CUDA 13.3, RTX PRO 6000 Blackwell (sm_120), DeepGEMM `891d57b4` (vLLM default).

eugr:
@ormandj - you can revert to old behavior by setting `VLLM_USE_DEEP_GEMM=0` environment variable.

arthur-drozdov:
Tested the merged `nv_dev` SM120 support on dual DGX Spark GB10 / SM121 through the vLLM DeepSeek-V4-Flash path. The package installs and vLLM can see DeepGEMM, but the current DeepGEMM scale-layout path still does not boot for this model stack.

Build/runtime:

- vLLM `sm120-pr-41834-stable-preview-20260624` / `28fef2c703` (`20260625.dev0+g28fef2c70.d20260625`)
- DeepGEMM `nv_dev` / `2073ddb2814892014c33ef4cd1c7d4c148baf1fe` (`deep-gemm 2.5.0+2073ddb`)
- PyTorch `2.11.0+cu130`, FlashInfer `0.6.13`, NCCL `2.30.7`
- `deepseek-ai/DeepSeek-V4-Flash`, TP=2, CUDA graphs on, `--load-format safetensors`

What fails:

1. Full DeepGEMM enabled:

```text
VLLM_USE_DEEP_GEMM=1
VLLM_MOE_USE_DEEP_GEMM=1
VLLM_USE_DEEP_GEMM_E8M0=1
```

The engine fails during weight post-processing before server startup:

```text
vllm/model_executor/kernels/linear/scaled_mm/deep_gemm.py
  -> deepgemm_post_process_fp8_weight_block
  -> deepgemm_post_process_weight_scale_block
  -> transform_sf_into_required_layout
RuntimeError: CUDA driver error (.../jit/handle.hpp:141): 200 (CUDA_ERROR_INVALID_IMAGE, device kernel image is invalid)
```

2. Tried isolating DeepGEMM to MoE only by keeping linear on Triton:

```text
--linear-backend triton
--moe-backend deep_gemm
VLLM_USE_DEEP_GEMM=1
VLLM_MOE_USE_DEEP_GEMM=1
```

This selects the intended split:

```text
Selected TritonFp8BlockScaledMMKernel for Fp8LinearMethod
Using 'DEEPGEMM_MXFP4' Mxfp4 MoE backend.
```

but still fails in MXFP4 MoE scale packing:

- with `VLLM_USE_DEEP_GEMM_E8M0=0`:

```text
RuntimeError: Assertion error (csrc/apis/layout.hpp:49): not disable_ue8m0_cast
_pack_deepgemm_mxfp4_scales -> transform_sf_into_required_layout
```

- with `VLLM_USE_DEEP_GEMM_E8M0=1`:

```text
RuntimeError: CUDA driver error (.../jit/handle.hpp:141): 200 (CUDA_ERROR_INVALID_IMAGE, device kernel image is invalid)
_pack_deepgemm_mxfp4_scales -> transform_sf_into_required_layout
```

Working workaround for this stack is still to disable DeepGEMM globally:

```text
VLLM_USE_DEEP_GEMM=0
VLLM_USE_DEEP_GEMM_E8M0=0
```

That boots cleanly with:

```text
Selected TritonFp8BlockScaledMMKernel for Fp8LinearMethod
Using 'MARLIN' Mxfp4 MoE backend.
DeepSeek V4: using official FlashInfer SM120 packed sparse-MLA decode
Application startup complete
```

So this is not just a vLLM linear-backend selection issue. On GB10, the `DEEPGEMM_MXFP4` scale packing path also reaches the same invalid-image class of failure. Happy to rerun with extra `DG_JIT_*` diagnostics if there is a specific knob or branch worth trying.

leavelet:
> Tested the merged `nv_dev` SM120 support on dual DGX Spark GB10 / SM121 through the vLLM DeepSeek-V4-Flash path. The package installs and vLLM can see DeepGEMM, but the current DeepGEMM scale-layout path still does not boot for this model stack.
> 
> Build/runtime:
> 
> * vLLM `sm120-pr-41834-stable-preview-20260624` / `28fef2c703` (`20260625.dev0+g28fef2c70.d20260625`)
> * DeepGEMM `nv_dev` / `2073ddb2814892014c33ef4cd1c7d4c148baf1fe` (`deep-gemm 2.5.0+2073ddb`)
> * PyTorch `2.11.0+cu130`, FlashInfer `0.6.13`, NCCL `2.30.7`
> * `deepseek-ai/DeepSeek-V4-Flash`, TP=2, CUDA graphs on, `--load-format safetensors`
> 
> What fails:
> 
> 1. Full DeepGEMM enabled:
> 
> ```
> VLLM_USE_DEEP_GEMM=1
> VLLM_MOE_USE_DEEP_GEMM=1
> VLLM_USE_DEEP_GEMM_E8M0=1
> ```
> 
> The engine fails during weight post-processing before server startup:
> 
> ```
> vllm/model_executor/kernels/linear/scaled_mm/deep_gemm.py
>   -> deepgemm_post_process_fp8_weight_block
>   -> deepgemm_post_process_weight_scale_block
>   -> transform_sf_into_required_layout
> RuntimeError: CUDA driver error (.../jit/handle.hpp:141): 200 (CUDA_ERROR_INVALID_IMAGE, device kernel image is invalid)
> ```
> 
> 2. Tried isolating DeepGEMM to MoE only by keeping linear on Triton:
> 
> ```
> --linear-backend triton
> --moe-backend deep_gemm
> VLLM_USE_DEEP_GEMM=1
> VLLM_MOE_USE_DEEP_GEMM=1
> ```
> 
> This selects the intended split:
> 
> ```
> Selected TritonFp8BlockScaledMMKernel for Fp8LinearMethod
> Using 'DEEPGEMM_MXFP4' Mxfp4 MoE backend.
> ```
> 
> but still fails in MXFP4 MoE scale packing:
> 
> * with `VLLM_USE_DEEP_GEMM_E8M0=0`:
> 
> ```
> RuntimeError: Assertion error (csrc/apis/layout.hpp:49): not disable_ue8m0_cast
> _pack_deepgemm_mxfp4_scales -> transform_sf_into_required_layout
> ```
> 
> * with `VLLM_USE_DEEP_GEMM_E8M0=1`:
> 
> ```
> RuntimeError: CUDA driver error (.../jit/handle.hpp:141): 200 (CUDA_ERROR_INVALID_IMAGE, device kernel image is invalid)
> _pack_deepgemm_mxfp4_scales -> transform_sf_into_required_layout
> ```
> 
> Working workaround for this stack is still to disable DeepGEMM globally:
> 
> ```
> VLLM_USE_DEEP_GEMM=0
> VLLM_USE_DEEP_GEMM_E8M0=0
> ```
> 
> That boots cleanly with:
> 
> ```
> Selected TritonFp8BlockScaledMMKernel for Fp8LinearMethod
> Using 'MARLIN' Mxfp4 MoE backend.
> DeepSeek V4: using official FlashInfer SM120 packed sparse-MLA decode
> Application startup complete
> ```
> 
> So this is not just a vLLM linear-backend selection issue. On GB10, the `DEEPGEMM_MXFP4` scale packing path also reaches the same invalid-image class of failure. Happy to rerun with extra `DG_JIT_*` diagnostics if there is a specific knob or branch worth trying.

@arthur-drozdov vllm mainline has already merged sm120 support—please use mainline vllm. jasl's PR is out of support scope. Please see https://github.com/vllm-project/vllm/pull/43477 for reference

PR #346, opened 2026-05-31
fix: use dynamic shared memory in smxx_paged_mqa_logits_metadata to support large batch sizes

### LINKED ISSUE #322: `get_paged_mqa_logits_metadata()` does not work with large batch size due to excessive static shared memory impl
```python
import torch

from vllm.third_party.deep_gemm import get_paged_mqa_logits_metadata

batch_size = 8192
block_size = 64
num_sms = torch.cuda.get_device_properties(0).multi_processor_count

context_lens = torch.full((batch_size, 1), 4096, dtype=torch.int32, device="cuda")
indices = torch.arange(batch_size, dtype=torch.int32, device="cuda") // 2

get_paged_mqa_logits_metadata(
    context_lens,
    block_size,
    num_sms,
    indices=indices,
)
torch.cuda.synchronize()
```
```
NVCC compilation failed: ptxas error   : Entry function '_ZN9deep_gemm5sched30smxx_paged_mqa_logits_metadataILj8192ELj256ELj152ELb1EEEvjjbPKjS3_Pj' uses too much shared data (0x18004 bytes, 0xc000 max)

  File "/home/thien/debug.py", line 12, in <module>
    get_paged_mqa_logits_metadata(
RuntimeError: Assertion error (/workspace/.deps/deepgemm-src/csrc/apis/../jit_kernels/impls/../../jit/compiler.hpp:228): false and "NVCC compilation failed"
```

This is caused by static shared memory being linear with batch size

https://github.com/deepseek-ai/DeepGEMM/blob/891d57b4db1071624b5c8fa0d1e51cb317fa709f/deep_gemm/include/deep_gemm/scheduler/paged_mqa_logits.cuh#L19-L20

Would it possible to make it support large batch size instead? Either with dynamic shared memory or reimplement in a way that doesn't need that much shared memory.

The context is that I'm trying to use `fp8_fp4_paged_mqa_logits()` for prefill case, which requires calling this metadata preparation kernel. Thank you.

Oxygen56:
## Summary

Fixes #322. The `smxx_paged_mqa_logits_metadata` kernel used static `__shared__` arrays sized by `kAlignedBatchSize`, which caused shared memory overflow at batch sizes >= 4096. The three arrays (`varlen_atom_token_start`, `varlen_atom_context_len`, `prefix_sum`) consumed `3 * kAlignedBatchSize * 4` bytes, exceeding the 48KB limit.

## Changes

1. **`deep_gemm/include/deep_gemm/scheduler/paged_mqa_logits.cuh`**: Replaced three static `__shared__` arrays with an `extern __shared__` pool, using pointer arithmetic to partition it. Updated `varlen_num_atoms_shared` accesses to use pointer dereferencing.

2. **`csrc/jit_kernels/impls/smxx_fp8_paged_mqa_logits.hpp`**: Fixed the `smem_size` calculation from `aligned_batch_size * sizeof(int)` to `(3 * aligned_batch_size + 1) * sizeof(int)` to match the actual allocation (the FP4 version already had the correct calculation).

## Background

The dynamic shared memory approach works because the `construct_launch_config` function in the JIT runtime already calls `cudaFuncSetAttribute(..., cudaFuncAttributeMaxDynamicSharedMemorySize, smem_size)`, which raises the per-block shared memory limit beyond the default 48KB to support the requested allocation (up to the GPU's capacity of ~227KB).

Tested with the reproducer from the issue at batch_size=8192.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/csrc/jit_kernels/impls/smxx_fp8_paged_mqa_logits.hpp
+++ b/csrc/jit_kernels/impls/smxx_fp8_paged_mqa_logits.hpp
@@ -61,8 +61,8 @@ static void smxx_paged_mqa_logits_metadata(const torch::Tensor& context_lens,
     const int aligned_batch_size = align(batch_size, 32);
     const int split_kv = block_kv * num_math_warpgroups;
     
-    // Calculate shared memory size
-    const int smem_size = aligned_batch_size * static_cast<int>(sizeof(int));
+    // Shared memory: prefix_sum[kAlignedBatchSize] + varlen_atom_token_start/context_len[kAlignedBatchSize] + varlen_num_atoms
+    const int smem_size = (3 * aligned_batch_size + 1) * static_cast<int>(sizeof(int));
     DG_HOST_ASSERT(smem_size <= SM90ArchSpec::smem_capacity);
     DG_HOST_ASSERT(smem_size <= SM100ArchSpec::smem_capacity);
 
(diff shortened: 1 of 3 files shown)


zheanxu:
Fixed in https://github.com/deepseek-ai/DeepGEMM/pull/347

PR #379, opened 2026-07-15
SM120: fix FP8 MQA logits swizzle mode for head_dim < 128

leavelet:
kSwizzleMode was hardcoded to 128 while the TMA descriptor uses swizzle_mode = head_dim (and kSwizzleAlignment = head_dim * 8 = the swizzle span). For head_dim in {32, 64} the kernel read SMEM with a 128B swizzle over a 32/64B-swizzled layout, mis-addressing reads into other (valid, already-written) pipeline stages: racecheck-clean but timing-dependent, so logits were both wrong and nondeterministic at deep KV pipelines. Set kSwizzleMode = kHeadDim (SwizzleContext supports B32/B64/B128). Same fix applied to the paged kernel, where head_dim is currently always 128 (no behavior change).

Repro on RTX PRO 6000 (head_dim 32/64, seq_len_kv 8192): diff vs reference ~0.98 and nondeterministic across runs; after the fix all head_dim in {32, 64, 128} are deterministic with diff ~1e-6.

Also adds FP8 head_dim 32/64 coverage to test_mqa_logits (previously only head_dim=128 was exercised, which is why this never fired).

### DIFF
--- a/deep_gemm/include/deep_gemm/impls/sm120_fp8_paged_mqa_logits.cuh
+++ b/deep_gemm/include/deep_gemm/impls/sm120_fp8_paged_mqa_logits.cuh
@@ -53,7 +53,8 @@ void sm120_fp8_paged_mqa_logits(const uint32_t batch_size,
     static constexpr uint32_t kNumMathWarps = kNumMathThreads / 32;
     static constexpr uint32_t kWarpsPerGroup = BLOCK_KV / MMA_M;
     static constexpr uint32_t kNumGroups = kNumMathWarps / kWarpsPerGroup;
-    static constexpr uint32_t kSwizzleMode = 128;
+    // Must match the TMA descriptor's swizzle (host passes swizzle_mode = head_dim)
+    static constexpr uint32_t kSwizzleMode = kHeadDim;
     static constexpr uint32_t kSMEMKBytes = kHeadDim;
 
     DG_STATIC_ASSERT(kNumTMAThreads == 128, "Expected 128 TMA threads");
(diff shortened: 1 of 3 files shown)


leavelet:
@RayWang96 Please review, thanks

deep_gemm/include/deep_gemm/impls/sm120_fp8_paged_mqa_logits.cuh
```diff
@@ -53,7 +53,8 @@ void sm120_fp8_paged_mqa_logits(const uint32_t batch_size,
     static constexpr uint32_t kNumMathWarps = kNumMathThreads / 32;
     static constexpr uint32_t kWarpsPerGroup = BLOCK_KV / MMA_M;
     static constexpr uint32_t kNumGroups = kNumMathWarps / kWarpsPerGroup;
-    static constexpr uint32_t kSwizzleMode = 128;
+    // Must match the TMA descriptor's swizzle (host passes swizzle_mode = head_dim)
+    static constexpr uint32_t kSwizzleMode = kHeadDim;
```
bvolpato, line 57: Could we add focused paged FP8 coverage for `head_dim` 32 and 64? The public paged API accepts both dimensions, but `test_paged_mqa_logits` only exercises 128. Is the missing coverage a concern here, given this swizzle bug was timing-dependent?

bvolpato (commented):
Other than the comment below, change looks good.

aganhui:
Independent confirmation of this fix on an RTX 5090 D (SM120), nv_dev @ `2642b32`, CUDA 13.0, PyTorch 2.11.0+cu130:

- **384-config self-consistency scan** over every enumerated `test_mqa_logits` case (arch 12): **189 failures without the fix** — all fp8 with head_dim 32/64, zero at D=128, zero mxfp4. **0/384 with this branch applied.**
- Full stock `test_mqa_logits` passes with the change.
- One behavior worth flagging for review context: the corruption has **two distinct manifestations**. In the dense kernel the misreads hit uninitialized shared memory → in-window logits are nondeterministic run-to-run (self-consistency catches it). In the **paged** kernel the misreads hit deterministically-stale memory → the wrong values are bit-stable across runs, **self-consistency passes**, and only the accuracy comparison vs the test's reference exposes it (fp8 paged D=32/64 diff was 0.83/0.72 before, ~1e-3 with this fix). This is likely also the root of the SM120 fp8 symptoms reported in #392 in July.
- Evidence and the scan script: https://gist.github.com/aganhui/f783450b212217284c062c3b46300649

We independently re-derived the same fix in #434 before noticing this PR — apologies for the duplication, and happy for #434 to be closed in favor of this one. One small addition you may want to fold in here: a `DG_STATIC_ASSERT(kHeadDim == 32 or kHeadDim == 64 or kHeadDim == 128)` next to the derived constant, so a future dispatch widening cannot silently instantiate an invalid `CuTeSwizzle`.

Issue #333, opened 2026-05-12
[Feature Request] Make binary artifacts portable on torch and Python versions


Harry-Chen:
Currently this project uses pybind11, which is coupled with both torch and Python ABIs. This leads to a huge amount of release artifacts (M * N, now 86) for each version, and sometimes causes some trouble to downstream projects that want to vendor it (https://github.com/vllm-project/vllm/pull/41516, https://github.com/vllm-project/vllm/issues/41487).

To make it portable, my suggestions are:

1. Switch to https://github.com/apache/tvm-ffi as a whole. This might not be easy since this project depends on torch a lot. Or,
2. Provides ABI stability on both torch and Python by migrating to [LibTorch Stable ABI](https://docs.pytorch.org/docs/stable/notes/libtorch_stable_abi.html) and compile the Cpp interface with [`Py_LIMITED_API`](https://docs.python.org/3/c-api/stable.html#limited-c-api). So we can have 1 binary wheel on each architecture that is compatible with Torch 2.9+ and Python 3.8+. Or,
3. Moving to `TORCH_LIBRARY` (not using stable ABIs) and opt-in`Py_LIMITED_API`. Then the wheel still depends on a specific torch version, but a wide range of Python versions from 3.8+. Or,
4. Replacing `pybind11` with `nanobind`. This requires minimal effort with Python stable ABI support, but only ranging from 3.12+.

I can help with any option if needed.

PR #388, opened 2026-07-20
Make the JIT cache key independent of the install include path

matteso1:
The kernel cache key embeds the absolute `-I{library_include_path}` token via `flags`, so a byte-identical kernel re-JITs and duplicates cache entries whenever the install prefix differs (venv rename, container build-layer vs runtime layout, moved NFS mount). That defeats sharing a `DG_JIT_CACHE_DIR` across layouts.

This strips that one token from the key (`strip_flag_token`, exact match, key computation only; the compile command and `DG_JIT_DEBUG` output are unchanged). It stays safe because that directory's header content is already keyed through the include hash embedded in `code` (`IncludeParser::get_hash_value`); everything else stays keyed: the compiler-version `signature`, every behavioral flag, and every other `-I` (NVRTC's `cuda_home/include` headers are not content-hashed, so their path stays). Since the key changes, existing entries re-JIT once under it.

This PR follows simon-mo's request on the open vllm-project/vllm#48190 to propose the relocation fix upstream instead of carrying it as a vendored patch. The vendored variant proposed there also drops the version field; on H100, relocating a prebuilt cache across install layouts turned a guaranteed miss into a hit and cut 93.7s from a Qwen3-30B-A3B-FP8 cold start (n=1). This PR deliberately keeps the version field keyed.

Related but distinct: #301/#302 (JIT cache concurrency), #333 (wheel portability).


matteso1:
Closing and resubmitting this change so the current automated review runs, per maintainer request.

PR #398, opened 2026-08-05
Make the JIT cache key independent of the install include path

matteso1:
This resubmits #388 so the repository's current automated review runs.

## The bug

The kernel cache key embeds the absolute `-I{library_include_path}` token through `flags`. Byte-identical kernels therefore re-JIT and create duplicate cache entries when the install prefix changes, such as a renamed virtual environment or a container build path that differs from the runtime path. That prevents a shared `DG_JIT_CACHE_DIR` from being portable across those layouts.

This follows simon-mo's request on vllm-project/vllm#48190 to propose the relocation fix upstream rather than carry it as a vendored patch. Related but distinct work includes #301/#302 for JIT cache concurrency and #333 for wheel portability.

## The fix

Strip only the exact `-I{library_include_path}` token when computing the cache key. The actual compile command and `DG_JIT_DEBUG` output are unchanged.

To keep reuse safe across different installed bundles, the full installed include tree is hashed once per process by relative path and file content. That tree hash covers the packaged DeepGEMM, CUTE, and CUTLASS headers without depending on the install prefix. The compiler-version `signature`, every behavioral flag, and every other include path also remain keyed. Existing entries re-JIT once under the new key.

## Measured

The broader vendored variant in vllm-project/vllm#48190 also drops the compiler-version field. On one H100 NVL with Qwen3-30B-A3B-FP8, TP1, and DeepGEMM `a6b593d`, relocating a planted cache caused the stock key to recompile 41/41 kernels and start in 170.6 seconds. The tested portable key reused 41/41 and started in 76.9 seconds, a 93.7-second difference, n=1 per arm.

This PR is intentionally narrower because it retains compiler-version identity. Its exact diff has not been independently timed on GPU hardware.

## Test plan

- `git diff --check`
- A local CPU regression verified that identical installed header trees produce the same hash across two roots, while changing a bundled CUTE header changes the hash.
- Source review against current DeepGEMM `main` at `559d79fb6994a58b8a15b4b93bf13ccc16edf247`


csrc/jit/compiler.hpp
```diff
@@ -97,8 +97,31 @@ class Compiler {
...
+        return out;
+    }
+
     std::shared_ptr<KernelRuntime> build(const std::string& name, const std::string& code) const {
-        const auto kernel_signature = fmt::format("{}$${}$${}$${}", name, signature, flags, code);
+        // Exclude the library include path from the cache key: it is location identity,
+        // not kernel identity, as its headers' content is already hashed into `code`.
+        // Every other flag stays keyed, including other `-I` paths (e.g. the CUDA home
+        // include, whose headers are not content-hashed) and the compiler `signature`.
+        const auto key_flags = strip_flag_token(flags, "-I" + library_include_path.string());
```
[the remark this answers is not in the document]
matteso1: Verified this against current `main`. `setup.py` copies CUTE and CUTLASS into the installed include tree, while `IncludeParser` only followed `deep_gemm/*`, so the old diff could reuse a CUBIN across different bundled headers.

Fixed in `06a6ae2` by hashing every installed include file by relative path and content once per process, then including that tree hash in the generated-code hash. A local CPU regression now verifies that relocating identical trees is a hit and changing a bundled CUTE header is a miss. I also corrected the token-removal comment.

csrc/jit/include_parser.hpp
```diff
@@ -12,6 +17,30 @@ namespace deep_gemm {
...
     std::unordered_map<std::string, std::optional<std::string>> cache;
+    std::optional<std::string> library_hash;
+
+    std::string get_library_hash_value() {
+        if (library_hash.has_value())
+            return library_hash.value();
+
+        // Key the installed header tree by relative path and content, not its install prefix.
+        std::vector<std::filesystem::path> files;
+        for (const auto& entry: std::filesystem::recursive_directory_iterator(library_include_path)) {
```
bvolpato, line 28: Could this reintroduce the `std::filesystem::recursive_directory_iterator` ABI breakage addressed in `436a563`? That change replaced this iterator with recursive `directory_iterator` traversal after it broke shipped wheels. Is the same compatibility concern applicable here?

bvolpato, line 28: `develop.sh` installs `cute` and `cutlass` as directory symlinks, while `recursive_directory_iterator` does not follow directory symlinks by default. Could that omit bundled headers from this hash and let source builds with different CUTE/CUTLASS revisions share a cache key and load a stale CUBIN?

PR #399, opened 2026-08-05
fix: SM90 paged mqa `prefix_sum` out of range (`CUDBG_EXCEPTION_WARP_OUT_OF_RANGE_ADDRESS`)

cjackal:
# Description

In `sm90_paged_mqa_logits.cuh`, [we compute `num_segs` using diff of `prefix_sum`](https://github.com/deepseek-ai/DeepGEMM/blob/559d79fb6994a58b8a15b4b93bf13ccc16edf247/deep_gemm/include/deep_gemm/scheduler/sm90_paged_mqa_logits.cuh#L113), but current diff calculation logic is unsafe as `q_idx` can be `batch_size` after the binary search which is out of bound of `prefix_sum` if `batch_size` is aligned. In this case, we should instead refer to the already computed segment sum (the would-be `prefix_sum[batch_size]`).

The overall math is trivial, so no further comments added in this ternary.

We have hit this error in production, the following is the CUDA coredump - `sm90_paged_mqa_logits_metadata<kAlignedBatchSize=96, SPLIT_KV=256, kNumSMs=132, kIsVarlen=false>` hits the problematic line with 16 sequences with next_n=6:

```log
...
[19:25:48.404432] coredump: Detected an exception of type CUDBG_EXCEPTION_WARP_OUT_OF_RANGE_ADDRESS (5)
[19:25:48.404435] coredump:   - Device: 4
[19:25:48.404437] coredump:   - SM: 124
[19:25:48.404439] coredump:   - Warp: 1
[19:25:48.404440] coredump:   - PC 0x7ef928bc46b0
[19:25:48.404688] coredump: Stack trace (lane masks: active 0xFFFFFFFF, valid 0xFFFFFFFF):
[19:25:48.404692] coredump:   # 0   0x7ef928bc46b0  _ZN9deep_gemm5sched30sm90_paged_mqa_logits_metadataILj96ELj256ELj132ELb0EEEvjjbPKjS3_Pj
...
```


cjackal:
It seems essentially the same bugfix has been merged on `nv_dev` branch by #353

PR #407, opened 2026-08-14
fix(layout): validate SF tensors are CUDA; add MXFP4 (1, 32) SF transform test

qqtang-code:
## Summary

Two small hardening changes for the scale-factor (SF) layout APIs, shared with the SGLang fork (sgl-project/DeepGEMM#75):

1. **`preprocess_sf` device validation** — the layout-transform functions feed `.data_ptr()` straight into JIT-launched device kernels (or into the `torch` fallback). A CPU tensor reaches the kernel launch with host pointers and segfaults the process (illegal device address) instead of failing cleanly. `preprocess_sf` now asserts `sf.is_cuda()` up front, covering both `get_mn_major_tma_aligned_tensor` and `get_mn_major_tma_aligned_packed_ue8m0_tensor`.

2. **Regression test for the MXFP4 `(1, 32)` recipe** — mirrors the exact call SGLang makes during Kimi-K3 / DeepSeek-V4 weight loading (`transform_sf_into_required_layout(sf, mn=..., k=..., recipe=(1, 32), num_groups=..., disable_ue8m0_cast=False)`), asserting the packed-UE8M0 output matches the torch reference (values, shape, strides). Skipped on SM90 (which uses the `(1, 128)` FP32 TMA layout) and on SM120 (upstream main has no arch-12 branch; SM120 support lives on the sgl-project fork).

## Context

The device assert turns a silent process crash (CPU tensor into a device kernel) into a clean assertion error. The regression test locks in the MXFP4 weight-prep path that SGLang's MoE runner relies on (validated end-to-end on 4× RTX 6000D / SM120 with the sgl fork: sgl-project/sglang#34827).

### DIFF
--- a/csrc/jit_kernels/impls/smxx_layout.hpp
+++ b/csrc/jit_kernels/impls/smxx_layout.hpp
@@ -110,6 +110,11 @@ static std::tuple<int, int, int, int, int, torch::Tensor> preprocess_sf(const to
     const auto dim = sf.dim();
     DG_HOST_ASSERT(dim == 2 or dim == 3);
     DG_HOST_ASSERT(sf.scalar_type() == torch::kFloat);
+    // The downstream paths launch device kernels over `sf` (or read
+    // `.data_ptr()` on the JIT launch path); a CPU tensor would crash the
+    // process at launch time (illegal device address) instead of failing
+    // cleanly. Validate the device up front.
+    DG_HOST_ASSERT(sf.is_cuda());
     const auto batched_sf = dim == 2 ? sf.unsqueeze(0) : sf;
 
     const auto [num_sf_batches, mn, sf_k] = get_shape<3>(batched_sf);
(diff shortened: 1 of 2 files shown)


qqtang-code:
Thanks for the thorough review — both suggestions addressed in 73ccae7:

1. **`preprocess_sf` comment** — you're right that `get_shape` (`csrc/utils/layout.hpp`) already asserts `t.is_cuda()` upstream, so a CPU tensor never actually reaches a kernel launch here. Kept the assert (it makes the SF-layout entry points fail earlier and more explicitly) but corrected the comment to say exactly that: early validation at the entry point, with `get_shape` re-validating downstream. Worth noting the sgl-project fork's `get_shape` **dropped** the `is_cuda` check during its SM120 work, so there this assert is the primary device guard — keeping the C++ hunk byte-identical across both repos (sgl-project/DeepGEMM#75) is intentional.

2. **Test print order** — moved the section title before the arch-skip, matching `test_k_grouped_psum_sf_layout_kernels`. Applied the same style to the fork's copy of the test.

Issue #417, opened 2026-08-24
[Regression] 8b1392b removes SM12x pure-fp8 1d1d kernels and aliases fp8_gemm_nt to the fp4 dispatcher (silent corruption on GB10)


maci0:
## Summary

Between `a6b593d` and `8b1392b` on `nv_dev`, the SM12x pure-fp8 path regressed.

Verified diff (`git diff a6b593d..8b1392b`):

- **Removed**: `csrc/jit_kernels/impls/sm100_fp8_gemm_1d1d.hpp` (−416) and
  `deep_gemm/include/deep_gemm/impls/sm100_fp8_gemm_1d1d.cuh` (−567) — the
  pure-fp8 1d1d kernels.
- `fp8_fp4_mqa_logits` dispatch rewritten: generic `smxx_fp8_mqa_logits`
  → per-arch `sm90_fp8_mqa_logits` / `sm100_mqa_logits` / `sm120_mqa_logits`.
- NOTE (correction of the earlier draft): the `fp8_gemm_nt = fp8_fp4_gemm_nt`
  alias is **not** new — it exists in `a6b593d` already (`gemm.hpp:792`).
  The regression is the 1d1d kernel removal, not the alias.

## Impact (measured on 2x DGX Spark, GB10 / sm_121a)

With `--linear-backend deep_gemm` (DeepSeek-V4-Flash FP8 linear + MLA),
fp8xfp8 inputs on SM12x run the combined `sm120_fp8_fp4_gemm_1d1d` kernel,
which misreads fp8 weights as fp4: **silent numerical corruption**. Greedy
France output degenerates (`' Septy Septy…'`) and DSpark draft acceptance
collapses, dropping decode from ~25.8 tok/s (b12x linear baseline) to
~4.4 tok/s.

`a6b593d` (the last good commit, frozen by eugr/spark-vllm-docker and used
by vLLM v0.25.1) serves correctly.

## Request

Restore the pure-fp8 1d1d kernels on `nv_dev` (or route fp8xfp8 on SM12x
back to a pure-fp8 kernel) so the fp8 path is correct again. vLLM's cmake
pin is currently `8b1392b`; we have opened a vLLM PR
([#53680](https://github.com/vllm-project/vllm/pull/53680)) to pin back to
`a6b593d` in the interim.

maci0:
[...]

Issue #372, opened 2026-07-06
transform_sf_into_required_layout: no SM120/121 (arch_major=12) branch for (gran_mn=1, gran_k=32) NVFP4 expert scales -- blocks DeepSeek-V4 weight loading


zbrad:
## Summary

`transform_sf_into_required_layout` in `csrc/apis/layout.hpp` has no branch for `arch_major=12` (SM120/SM121, i.e. RTX 5090 / RTX Pro 6000 Blackwell / NVIDIA GB10) at the `(gran_mn=1, gran_k=32)` scale-factor granularity DeepSeek-V4's NVFP4 expert weights use. This is a distinct kernel/dispatch gap from the ones already fixed in #317 / PR #318 (`tf32_hc_prenorm_gemm`, `paged_mqa_logits`) -- it blocks DeepSeek-V4 at **weight loading time** (`process_weights_after_loading`), before any forward pass is even attempted.

## Environment

- GPU: NVIDIA GB10 (DGX Spark), compute capability 12.1 (SM121)
- Framework: vLLM (`vllm-project/vllm`, personal GB10 fork), vendored DeepGEMM at `891d57b4` (2026-04-24)
- Model: `yujiepan/deepseek-v4-tiny-random` -- a small test model whose config faithfully mirrors `deepseek-ai/DeepSeek-V4-Flash`'s real quantization scheme (`expert_dtype: fp4`, same `quantization_config`), used here specifically to reproduce this on constrained hardware

## Exact failure

```
RuntimeError: Assertion error (csrc/apis/layout.hpp:59): Unknown SF transformation
```

Traceback (vLLM side):
```
File "vllm/model_executor/model_loader/utils.py", line 113, in process_weights_after_loading
File "vllm/model_executor/layers/quantization/fp8.py", line 444, in process_weights_after_loading
File "vllm/model_executor/kernels/linear/scaled_mm/deep_gemm.py", line 96, in process_weights_after_loading
File "vllm/model_executor/layers/quantization/utils/fp8_utils.py", line 1140, in deepgemm_post_process_fp8_weight_block
File "vllm/model_executor/layers/quantization/utils/fp8_utils.py", line 1079, in deepgemm_post_process_weight_scale_block
File "vllm/utils/deep_gemm.py", line 494, in transform_sf_into_required_layout
```

## Root cause analysis

`transform_sf_into_required_layout` (`csrc/apis/layout.hpp`) has exactly four branches:

```cpp
[...]

PR #419, opened 2026-08-26
[SM12x] Restore pure-fp8 1d1d path (removed in 8b1392b); fix kernel TU preamble

maci0:
## Summary

`8b1392b` deleted the pure-fp8 1d1d kernels (`sm100_fp8_gemm_1d1d.{hpp,cuh}`, −983 lines) and rewrote the SM12x fp8 dispatch. On arch 12, fp8xfp8 inputs then run the combined `sm120_fp8_fp4_gemm_1d1d` kernel, which misreads fp8 weights as fp4 — **silent numerical corruption** of FP8 linear layers.

Measured on 2x DGX Spark (GB10 / sm_121a), DeepSeek-V4-Flash-0731, vLLM `--linear-backend deep_gemm`:
- `8b1392b` pin: France output degenerates (`' Septy Septy…'`), DSpark draft acceptance collapses, decode ~25.8 → ~4.4 tok/s.
- `a6b593d` pin (last commit with the pure-fp8 kernel): coherent output.

Tracked in #417; vLLM-side interim pin: vllm-project/vllm#53680.

## What this restores

- `csrc/jit_kernels/impls/sm100_fp8_gemm_1d1d.hpp` (JIT kernel runtime)
- `deep_gemm/include/deep_gemm/dg25_fp8/` — the pure-fp8 kernel + its common/ dependencies as a self-contained module
- `gemm.hpp`: fp8xfp8 on arch 10/12 routes to `sm100_fp8_gemm_1d1d`; fp4 keeps the combined kernel (AB-swap paths handled).

## Header fixes the a6b593d TU needs to compile at all

The original 1d1d TU cannot be compiled as generated:

- `common/math.cuh` was never included by the 1d1d include chain → `align`/`swap` in `common/scheduler.cuh:192` / `common/sm100_utils.cuh:127` were undefined (`identifier "align" is undefined`). The TU now includes it and pulls in `deep_gemm::math` (include before the impl header, so the using-directive precedes the dependent code).
- `math.cuh`'s `cast_into_bf16_and_pack` needs `#include <cuda_bf16.h>` (`__float22bfloat162_rn`).
- `std::is_same_v` → `cute::is_same_v` for NVRTC (libcu++ does not expose variable templates in `std` under NVRTC).

## Validation status (honest)

- NVRTC frontend (CUDA 13.3): with the header fixes the TU compiles to PTX for `compute_121a` (~231 KB) — the preamble fix is verified.
- GB10 JIT toolchain gap (same as #417): stock CUDA 13.0/13.3 ptxas rejects `tcgen05.mma`/`.cta_group::1`/`.block32` for every sm_12x target; the NVRTC cubin is rejected by the driver (`CUDA_ERROR_INVALID_IMAGE`); the driver-JIT PTX route is rejected too (`CUDA_ERROR_INVALID_PTX`). So the kernel needs either a toolchain that can assemble tcgen05 for sm_121a, or build-time compilation (the golden/2.5.0 stack precompiles it).
- The `fp8_einsum` o_proj path runs correctly on SM121a with the restored stack (T=10/96/8192, finite output).

Happy to adjust the module naming (e.g. rename `dg25_fp8` → `sm100_fp8` or fold into the existing tree) to match maintainer preference.


maci0:
Thanks for the detailed review — all three criticals are addressed in 44d9d2e:

1. **SM120 swap_ab (transposed D)**: pure fp8 is now excluded from the host-side AB-swap decision (`swap_ab` requires `!pure_fp8`). The pure-fp8 branch always calls `sm100_fp8_gemm_1d1d` with the original a/b and m/n, with SF transformed unswapped — the restored kernel has no transposed-store support, and the SM100 heuristics cover small M with block_m<=32 candidates.

2. **SM100 swap layouts from heuristics**: added `GemmDesc::allow_swap_ab` (default `true`, so all existing kernels keep their candidate set); `SM100ArchSpec::get_layout_candidates` skips the `swap_ab==1` iteration when it is false; the restored runtime sets it false. This prevents both the silent store/smem mismatch and the `kIsMulticastOnA` static-assert JIT failure.

3. **arch-10 mixed FP4×FP8 routing**: the pure-fp8 branch now requires `a` *and* `b` to be non-FP4 (`a.first.scalar_type() != kPackedFP4 and b.first.scalar_type() != kPackedFP4`), matching the SM120 branch — fp4_A x fp8_B stays on the combined kernel.

Also folded in the minors: `epilogue_type` is wired through `get_default_epilogue_type(args.epilogue_type)` instead of the hardcoded literal; `dg25_fp8/common/math.cuh` now includes `cuda_bf16.h` itself (self-contained header); noted the `SM100ArchSpec::smem_capacity` reuse for sm_12x (GB10 matches SM100's 227 KB dynamic smem, so stage sizing is safe there).

Happy to re-run anything or adjust the `allow_swap_ab` approach if you'd prefer a different mechanism for filtering swap candidates.

maci0:
The re-review comments duplicate the earlier v4p/v5/v6 findings — all of them were addressed in 44d9d2e (the current head, plus the ci: re-trigger commit on top). Point-by-point:

1. **AB-swap for pure fp8 (gemm.hpp, `fp8_fp4_gemm_nt_sm120`)** — resolved: `const bool pure_fp8 = (a_data.scalar_type() != kPackedFP4) and (b_data.scalar_type() != kPackedFP4)` and the swap condition now includes `!pure_fp8`, so the restored kernel only ever runs non-swapped.

2. **`get_best_config<SM100ArchSpec>` may select swap_ab=true** (sm100_fp8_gemm_1d1d.hpp / heuristics) — resolved: `GemmDesc.allow_swap_ab=false` (config.hpp) is set in the restored kernel runtime, and `get_layout_candidates` (sm100.hpp:52) does `if (swap_ab == 1 and not desc.allow_swap_ab) continue;` — swapped layouts are never enumerated. The multicast-on-A hazard from the swapped `cluster_n=2` path can no longer be reached; with swap_ab=0, `cluster_n>1` is also filtered (sm100.hpp:93), so `kIsMulticastOnA` stays false and the `kNumMulticast == 1` static assert is not tripped.

3. **arch-10 routing checks only B's dtype** (gemm.hpp:218) — resolved: the pure-fp8 route condition now checks both operands (`a_data.scalar_type() != kPackedFP4 and b_data.scalar_type() != kPackedFP4`), so fp4-A × fp8-B stays on the combined kernel.

4. **sm_12x smem capacity** — comment added in the kernel runtime: GB10 per-SM dynamic smem matches SM100 (227 KB), so SM100ArchSpec stage sizing is safe on sm_12x. (Runtime verification on GB10 is pending a toolchain-capable wheel; the kernel JIT path itself compiles.)

5. **EpilogueIdentity hardcoded** — resolved: `get_default_epilogue_type(args.epilogue_type)` is wired back in the template args.

6. **math.cuh not self-contained** — resolved: `#include <cuda_bf16.h>` added at the top of `dg25_fp8/common/math.cuh`.

Happy to adjust anything that still looks off on re-review.

lucifer1004:
SM12x SMEM capacity is 100 KiB only.

maci0:
Related upstream scale-path work in the same area, for cross-reference:

- #372 / #403 — `transform_sf_into_required_layout` has no arch-12 branch for (gran_mn=1, gran_k=32) NVFP4 expert scales (blocks DSV4 weight loading on SM120/121); #403 allows explicit SM120 recipes to use the SM100 UE8M0 packing/validation.
- #337 — `transpose_and_pack_fp32_into_ue8m0` leaks mantissa bits into adjacent exponent bytes when packing non-power-of-2 FP32 scales.

This restore (pure-fp8 1d1d) and those two are complementary: kernel path restore + packer/transform correctness on SM12x. Worth reviewing together.

maci0:
Thanks for the catch — fixed in 54d5a3e.

`SM100ArchSpec::smem_capacity` (232448 = 227 KiB) was hardcoded and reused for sm_12x; with GB10 at only 100 KiB of opt-in per-block SMEM, the restored 1d1d kernel would over-allocate stages. `get_pipeline_config` now sizes stages from the device runtime's `sharedMemPerBlockOptin`, which is correct on both SM100 and SM12x, and the misleading "GB10 matches SM100" comment in the kernel runtime is corrected.

maci0:
Reproducible recipe: https://github.com/maci0/vllm-spark-0731 — the SM12x findings live in `docs/knowledge/09-golden-deepgemm.md` and `docs/UPSTREAM.md` (DeepGEMM section); `deepgemm-fp8-1d1d-port.diff` is our local pre-merge port of this restore.

Issue #414, opened 2026-08-24
[Bug] `pack_ue8m0_to_int` is not CUDA Graph capture safe


foraxe:
## Summary

`deep_gemm.utils.math.pack_ue8m0_to_int` performs CUDA-tensor value checks with
Python assertions:

```python
x_int = x.view(torch.int)
assert (x_int >= 0).all() and (x_int & 0x7FFFFF == 0).all()
```

Evaluating these zero-dimensional CUDA tensors as Python booleans forces a
device-to-host synchronization. When the helper runs during CUDA Graph capture,
that synchronization fails with `cudaErrorStreamCaptureUnsupported`.

## Affected integration

FlashInfer MegaMoE block-FP8 input staging calls this helper while SGLang
captures its decode graph. SGLang currently needs a narrowly scoped workaround
that replaces the helper only around the captured MegaMoE forward and restores
the original function immediately afterward:

- https://github.com/wenscarl/sglang/pull/6

## Expected behavior

DeepGEMM should expose a capture-safe UE8M0 packing path so downstream projects
do not need to replace a module-level helper. The dtype and shape checks can
remain host-side; the CUDA value checks could be optional, moved to an eager
validation path, or implemented without a host synchronization during capture.

The packing operation itself is capture safe:

```python
(x.view(torch.int) >> 23).to(torch.uint8).view(torch.int)
```

## Suggested validation

- Compare eager and captured/replayed packed output.
- Cover valid positive, zero-mantissa UE8M0 inputs.
- Preserve eager validation for malformed values if that remains part of the
  public contract.

## Version

Observed against DeepGEMM `main` at
`559d79fb6994a58b8a15b4b93bf13ccc16edf247`.

PR #421, opened 2026-08-26
fix: make pack_ue8m0_to_int CUDA-graph capture safe (#414)

XFDG:
## Problem

`pack_ue8m0_to_int` called `.all()` on device tensors:

```python
assert (x.view(torch.int) >> 23 >= 0).all()  # device→host sync!
assert (x.view(torch.int) & 0x7FFFFF == 0).all()  # device→host sync!
```

Inside a CUDA graph capture region, `.all()` triggers a device→host sync which raises `cudaErrorStreamCaptureUnsupported`, breaking SGLang's decode graph via the MegaMoE FP8 staging path (issue #414).

## Fix

Use `torch.cuda.is_current_stream_capturing()` to gate validation:

- **Outside capture**: bit-level assertions run, catching non-negative sign bit and zero mantissa — direct callers get a loud error on bad input
- **Inside capture**: assertions skipped — no device→host sync, capture-safe

Bit-level check avoids float rounding edge cases (e.g. subnormals like 2^-126 that `x.round()` would incorrectly round to 0).

```python
if not torch.cuda.is_current_stream_capturing():
    x_bits = x.view(torch.int)
    assert ((x_bits >> 31) == 0).all(), "scale values must be non-negative"
    assert ((x_bits & 0x7FFFFF) == 0).all(), "scale values must have zero mantissa"
```

## Test (H200, CUDA 12.8)

| Case | Result |
|------|--------|
| Valid ue8m0 from `ceil_to_ue8m0` | ✅ PASS |
| Non-ue8m0 input (mantissa set) | ✅ assert fires |
| Negative input | ✅ assert fires |
| Padding value 1.0 (valid ue8m0) | ✅ PASS |
| CUDA graph capture + invalid input | ✅ no assert, no sync |
| CUDA graph capture + valid input | ✅ replay works |

Addresses ds-review-bot v6 concern: validation is preserved outside capture paths.

### DIFF
--- a/deep_gemm/utils/math.py
+++ b/deep_gemm/utils/math.py
@@ -30,8 +30,10 @@ def pack_ue8m0_to_int(x: torch.Tensor):
     # region we still validate them so direct callers get a loud error; during
     # capture we skip the check to avoid a device->host sync.
     if not torch.cuda.is_current_stream_capturing():
-        assert (x >= 0).all(), "pack_ue8m0_to_int: scale values must be non-negative"
-        assert (x == x.round()).all(), "pack_ue8m0_to_int: scale values must have zero mantissa"
+        # Validate ue8m0 invariants at the bit level: sign=0, mantissa=0.
+        x_bits = x.view(torch.int)
+        assert ((x_bits >> 31) == 0).all(), "pack_ue8m0_to_int: scale values must be non-negative"
+        assert ((x_bits & 0x7FFFFF) == 0).all(), "pack_ue8m0_to_int: scale values must have zero mantissa"
     return (x_int >> 23).to(torch.uint8).view(torch.int)
 
 
--- a/deep_gemm/utils/math.py
+++ b/deep_gemm/utils/math.py
@@ -17,20 +17,24 @@ def ceil_to_ue8m0(x: torch.Tensor):
 
 
 def pack_ue8m0_to_int(x: torch.Tensor):
-    # NOTE: keep this helper CUDA-graph-capture safe. Calling `.all()` on a
-    # device tensor forces a device->host sync, which raises
-    # cudaErrorStreamCaptureUnsupported when invoked inside a captured region
-    # (e.g. SGLang's decode graph via the MegaMoE FP8 staging path).
-    #
-    # Host-side structural checks only:
+    """Pack a float32 tensor of UE8M0 scale factors into int32 (4 bytes per int).
+
+    Precondition: every element must be a power-of-two with zero mantissa and
+    non-negative sign (i.e. produced by ``ceil_to_ue8m0`` or the literal 1.0
+    padding). Violating this precondition silently produces wrong packed values
+    because the downstream GEMM/MHA kernels only validate shape/dtype, not
+    numerical content.
+
+    CUDA-graph-capture safe: ``.all()`` triggers a device→host sync which raises
+    ``cudaErrorStreamCaptureUnsupported`` inside a capture region. We therefore
+    skip the value assertions during capture and run them otherwise so that
+    direct callers outside capture still get a loud error on bad input.
+    """
     assert x.dtype == torch.float and x.size(-1) % 4 == 0
     x_int = x.view(torch.int)
-    # Value invariants (non-negative exponent, zero mantissa) are guaranteed by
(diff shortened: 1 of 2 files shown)


XFDG:
Thanks for the detailed review! Here's what was addressed:

**v6 warning** (silent corruption on invalid input outside capture):
Added conditional bit-level validation using `torch.cuda.is_current_stream_capturing()`:
- Outside capture: asserts fire on bad input (sign bit ≠ 0 or mantissa bits ≠ 0)
- Inside capture: assertions skipped to avoid `cudaErrorStreamCaptureUnsupported`
- Used bit-level checks (`x_bits >> 31`, `x_bits & 0x7FFFFF`) instead of float-based comparison to avoid false failures on subnormals like 2^-126

**v4p suggestion** (inaccurate comment "kernel itself traps on malformed scales"):
Removed — downstream kernels (`sm100_mqa_logits`, `sm100_fp8_fp4_gemm_1d1d`) only validate shape/dtype, not numerical content.

**v5 suggestion** (document preconditions):
Added a docstring explicitly stating the UE8M0 precondition, why violations aren't caught downstream, and the capture-safe rationale.

Tested on H200 (CUDA 12.8):
- Valid `ceil_to_ue8m0` output: ✅ passes
- Non-UE8M0 input (mantissa set): ✅ asserts outside capture
- Negative input: ✅ asserts outside capture  
- CUDA graph capture + invalid input: ✅ no assert, no device→host sync
- CUDA graph capture + valid input: ✅ capture and replay work

200lz:
I independently reproduced #414 while investigating the SGLang MegaMoE workaround. Your current capture-aware guard fixes the failure for valid inputs — verified on an RTX 3060 with PyTorch 2.6.0 + CUDA 11.8 (eager warm-up, then `torch.cuda.graph` capture/replay).

I found two small follow-ups that may be useful:

1. Guard the capture query with `x.is_cuda`, so CPU inputs keep eager value validation without touching CUDA capture state (this also lets the existing `x_int` bit view be reused).
2. Regression coverage for CUDA graph capture/replay, including mutating the static input buffer before replay to verify that packing is actually recorded and recomputed by the graph. I also kept an eager malformed-input test (nonzero mantissa, negative sign) to ensure the existing validation behavior outside capture is preserved.

I prepared the changes as a single commit on top of your current head (ba14b91):
https://github.com/200lz/DeepGEMM/commit/763b73df764b0a53fdc06e567d27cf880bea673f

Local results:

```
valid eager == capture/replay (bit-exact): PASS
mutated static input replay: PASS
eager malformed-input rejection: PASS
CPU short-circuit (CUDA_VISIBLE_DEVICES=""): PASS
git diff --check: PASS
```

Happy for you to cherry-pick the commit, or I can open a PR against your branch if that is easier.

XFDG:
Thanks a lot @200lz — cherry-picked your commit `763b73df` as `2a5989a` (authorship preserved). Both changes are clear wins:

- The `x.is_cuda and ...` short-circuit means CPU inputs always run the value validation instead of needlessly querying capture state, so malformed CPU input still errors loudly.
- The new `tests/test_pack_ue8m0.py` is exactly the capture/replay coverage the review asked for. The mutate-static-buffer-then-replay check is a nice touch — it proves the pack is recorded and recomputed by the graph rather than constant-folded at capture time.

Verified on an H200 (torch 2.10 / CUDA): all four checks pass against the real `math.py`, and a negative control confirms the pre-fix (unguarded) body still fails capture with `cudaErrorStreamCaptureUnsupported` — so the test genuinely guards the fix.

```
Testing CUDA graph capture/replay:
 > Capture/replay matches eager (bit-exact)
 > Replay after static-input mutation matches eager (bit-exact)
Testing eager malformed-input rejection:
 > AssertionError raised as expected (nonzero mantissa)
 > AssertionError raised as expected (negative sign)
```

Appreciate the independent repro and the follow-up.

XFDG:
Closed the three remaining review threads after rechecking the current head `2a5989a` against each item: the stale warning is fixed by eager bit-level validation outside capture, the precondition is documented, the inaccurate downstream-kernel comment is removed, and the CUDA graph capture/replay plus malformed-input regressions are present. No additional code change was needed. Please take another look when convenient.

PR #428, opened 2026-09-07
fix: reject non-positive block size multiples

nisbenz:
## Summary

Reject zero and negative values passed to `set_block_size_multiple_of`.

These values are currently stored without validation and later used by the
SM90 and SM100 kernel-selection heuristics. A zero value can produce
modulo-by-zero or a candidate loop with a zero increment.

This change fails early with a `RuntimeError` and adds regression coverage for
both scalar and `(M, N)` inputs.

## Testing

- `python -m compileall -q tests/test_runtime.py`
- `git diff --check`


nisbenz:
can you re do the code review

Issue #392, opened 2026-07-26
[nv_dev] SM120: MQA logits kernels fail test_attention self-consistency (fp8: nondeterministic/race) and accuracy (mxfp4 H=64 D=128)


linjiapro:
### Summary

On SM120 (`nv_dev` @ `f8e8fb5`, deep_gemm 2.6.1), the MQA logits kernels fail the repo's own `tests/test_attention.py::test_mqa_logits`, in two independent ways:

1. **FP8 format: nondeterministic output (race).** The test's self-consistency check — the same kernel on identical input across 20 repeat launches — fails with bitwise-different results:
   ```
   AssertionError: bitwise mismatch (mqa logits self-consistency): num_bytes=1048576,
   num_mismatch=56571, first_byte=250, elem=125, coord=(0, 125), byte_in_elem=0,
   x_byte=0, y_byte=134, x_val=0.0, y_val=1.046875
   ```
2. **MXFP4 format (`H=64, D=128`): accuracy failure.** `AssertionError: Diff: 0.6187` vs the 0.02 tolerance, at full throughput (~776 TFLOPS on this part) — wrong numbers, not a launch failure.

This kernel family has needed per-arch synchronization fixes before — SM90 (#229), SM100 (#285, and `96fbced` "Fix SM100 FP16 MQA synchronization") — and SM120 appears to be missing its equivalent.

### Environment

- GPU: NVIDIA RTX PRO 6000 Blackwell Max-Q Workstation Edition (SM120), driver 580.126.09
- CUDA 13.0, nvcc from `/usr/local/cuda/bin`, host compiler g++-10 (system g++ 9.4 lacks C++20 and fails JIT outright — separate, expected)
- torch 2.11.0+cu130, Python 3.12
- deep_gemm 2.6.1 built from `nv_dev` @ `f8e8fb5` via `setup.py bdist_wheel` (CC=gcc-10 CXX=g++-10)

### Repro

```bash
python - <<'EOF'
import sys
sys.path.insert(0, '<DeepGEMM>/tests')
import test_attention as ta
# fp8-only: the race repros here; the full sweep fails earlier on mxfp4 accuracy
orig = ta.sample_mqa_cases
ta.sample_mqa_cases = lambda tag, cases: [c for c in orig(tag, cases) if c[0] == 'fp8']
ta.test_mqa_logits()
EOF
```

[...]

PR #434, opened 2026-09-10
fix: derive the SM120 fp8 mqa_logits swizzle mode from head_dim

### LINKED ISSUE #433: [nv_dev][SM120] fp8_fp4_mqa_logits fp8 path returns nondeterministic in-window wrong results for head_dim 32/64 — read-side swizzle mode doesn't match the TMA descriptors
The read side of the SM120 fp8 MQA-logits kernel un-swizzles shared memory with a hardcoded 128-byte pattern, while the TMA descriptors and the copy helper swizzle in head_dim-byte atoms (32B/64B/128B for D=32/64/128). For head_dim 32/64 the patterns do not match, every fragment load addresses outside its row, and the kernel returns in-window logits that are silently wrong and run-to-run nondeterministic. D=128 is the only head_dim where the hardcoded value matches, which is why the main DSA path works.

## Failure surface (384-config self-consistency scan, script in the gist)

Every enumerated `test_mqa_logits` config (arch 12), 20 repeat calls compared bitwise:

```
TOTAL=384 FAILS=189
  fmt:        all fp8 (0 mxfp4 failures)
  head_dim:   D=32: 94   D=64: 95   D=128: 0
  compressed: both (CMP=0: 95, CMP=1: 94)
```

The stock test fails at its first fp8 config (fp8/bf16/uncompressed+clean, SQ=128, SK=4096, H=16, D=32, CP=1) with `x_val=0.0, y_val=7.28125` at coord (0,125) — inside the row's `[0, 640)` window. The exact `0.0`s come from misaddressed reads of uninitialized shared memory (`relu(0)*w*sf`), not from missing writes.

## Root cause — three sites that must agree

```cpp
// host TMA descriptors (csrc/jit_kernels/impls/sm120_mqa_logits.hpp):
make_tma_2d_desc(q, head_dim, ..., swizzle_mode=head_dim);        // per-D  ✓
// TMA write side (kernel):
tma::copy<kHeadDim, BLOCK_Q * kNumHeads, kHeadDim>(...);           // per-D  ✓
// kernel read side (sm120_fp8_mqa_logits.cuh):
static constexpr uint32_t kSwizzleMode = 128;                      // hardcoded ✗
```

The fp8 row is head_dim bytes, so writes land swizzled in head_dim-byte atoms. `SwizzleContext<kSwizzleMode>` and the fragment loaders then un-swizzle with the 128B pattern; for D=32/64 the XOR bits touch addresses the write pattern never permuted, so reads cross row boundaries of the tile or fall into pipeline stages that have not been filled yet — hence nondeterministic content.

## Fix (one line) and validation

```cpp
- static constexpr uint32_t kSwizzleMode = 128;
+ static constexpr uint32_t kSwizzleMode = kHeadDim;
```

The fix direction is the reader side: the smem tile geometry is "one row = head_dim bytes", so the swizzle atom must equal the row width (a 128B atom would span four D=32 rows and change the TMA box geometry). All readers are parameterized on this constant, so one change covers them.

Validated on RTX 5090 D (CUDA 13.0, torch 2.11.0+cu130, nv_dev @ 2642b32):
- self-consistency scan: **189 → 0** failures
- full `test_mqa_logits` (20x bitwise self-consistency + accuracy vs reference + bench): **PASSED**, all 384 configs, fp8 up to 411 TFLOPS

## Why related kernels are unaffected

- `sm120_fp4_mqa_logits.cuh`: D=128 only (`DG_STATIC_ASSERT`), row = 64 bytes, constant 64 — matches.
- `sm120_fp8_paged_mqa_logits.cuh` / fp4 paged: dispatch asserts `head_dim == 128`; their constant 128 matches the only supported head_dim.
- Dense fp8 is the only instantiation where multiple row widths are allowed by the dispatch (`head_dim == 32 or 64 or 128`) and the constant diverges from them.

## Relation to #392

On our hardware fp8 D=128 passes self-consistency across all configs (20 reps, both compressed and uncompressed), so the out-of-window nondeterminism reported in #392 does not reproduce here — possibly driver/timing dependent (the kernel, clean kernel, launch path and test file are byte-identical between f8e8fb5 and 2642b32). The mxfp4 H=64 D=128 accuracy failure from #392 also no longer reproduces. The stock test at f8e8fb5 already enumerated fp8 D=32/64, so the SM120 test failures reported there very likely included the defect fixed here.

## Scope / timing

`main` is unaffected (no SM120 support yet). When nv_dev next syncs to main, this would otherwise ship with the first official SM120 release: any fp8 DSA/MQA geometry with head_dim 32/64 on 5090-class hardware would silently corrupt logits.

Scan script and raw logs: https://gist.github.com/aganhui/f783450b212217284c062c3b46300649

PR incoming (base nv_dev). Happy to run further validation on this 5090 D if useful.

aganhui:
Fixes #433.

The SM120 fp8 MQA-logits readers un-swizzle shared memory with a hardcoded 128-byte pattern, while the host TMA descriptors and the `tma::copy` sites swizzle in head_dim-byte atoms (32B/64B/128B for D=32/64/128 — the fp8 row is head_dim bytes). For head_dim 32/64 the XOR bits don't match the write pattern, fragment loads address outside their row, and the kernel returns in-window logits that are silently wrong. D=128 is the only head_dim where the hardcoded value matches, which is why the main DSA path works.

This derives the mode from `kHeadDim`, matching the sibling kernels: SM90 (`to_swizzle_cute_type<kHeadDim>()`) and SM100 (`kHeadDim / kPackFactor`).

**Both fp8 variants are fixed** (second commit, addressing the review finding): the fp8 paged dispatch has no `head_dim == 128` restriction — only its fp4 sibling does — so the paged kernel had the same mismatch. Notably, in the paged variant the resulting garbage is *deterministic* (misreads land on deterministically-stale shared memory), so self-consistency passes and only the accuracy comparison against the reference exposes it.

Validation on RTX 5090 D (CUDA 13.0, PyTorch 2.11.0+cu130, nv_dev @ 2642b32):

- Dense, 384-config self-consistency scan: **189 failures → 0** (all failures were fp8 with D ∈ {32, 64}; zero at D=128, zero mxfp4)
- Dense, full `test_mqa_logits` (20x bitwise + accuracy vs reference + bench): **PASSED**, all 384 configs, fp8 up to 411 TFLOPS
- Paged, targeted fp8 repro (small configs; the full paged suite does not fit in 32 GB): accuracy diff vs reference **D=32: 0.83 → 0.0008, D=64: 0.72 → 0.001**, D=128 unchanged (~0.001)
- Dense regression after adding the static asserts: fp8 D=32 accuracy diff 0.0006 (< 1e-3)

Also adds `DG_STATIC_ASSERT(kHeadDim ∈ {32, 64, 128})` next to the derived constants (per review suggestion), so a future dispatch widening cannot silently instantiate an invalid `CuTeSwizzle`.

Unaffected kernels (verified): dense fp4 and both fp4 paged variants are head_dim-128-only (`DG_HOST_ASSERT`/`DG_STATIC_ASSERT`), where their constants already match. Dense fp8 and fp8 paged are the only instantiations whose dispatches allow multiple row widths, and both are now derived.

Scan script and raw logs: https://gist.github.com/aganhui/f783450b212217284c062c3b46300649

### DIFF
--- a/deep_gemm/include/deep_gemm/impls/sm120_fp8_mqa_logits.cuh
+++ b/deep_gemm/include/deep_gemm/impls/sm120_fp8_mqa_logits.cuh
@@ -58,7 +58,12 @@ void sm120_fp8_mqa_logits(const uint32_t seq_len, const uint32_t seq_len_kv,
     DG_STATIC_ASSERT(BLOCK_KV == kNumMathWarps * MMA_M, "BLOCK_KV = warps × MMA_M");
     DG_STATIC_ASSERT(kHeadDim % MMA_K == 0 and kNumHeads % MMA_N == 0, "Alignment");
 
-    static constexpr uint32_t kSwizzleMode = 128;
+    // Must match the host-side TMA descriptors (swizzle_mode = head_dim bytes) and
+    // the `tma::copy<..., kHeadDim>` smem layout: the fp8 row is head_dim bytes, so
+    // the TMA writes are swizzled in head_dim-byte atoms (32B/64B/128B for D=32/64/128).
+    // A hardcoded 128 here only matches D=128; for D=32/64 the readers would un-swizzle
+    // with the wrong pattern and consume out-of-tile shared memory.
+    static constexpr uint32_t kSwizzleMode = kHeadDim;
     static constexpr uint32_t kSwizzleAlignment = kHeadDim * 8;
     static constexpr uint32_t kSMEMKBytes = kHeadDim;
 
--- a/deep_gemm/include/deep_gemm/impls/sm120_fp8_mqa_logits.cuh
+++ b/deep_gemm/include/deep_gemm/impls/sm120_fp8_mqa_logits.cuh
@@ -64,6 +64,8 @@ void sm120_fp8_mqa_logits(const uint32_t seq_len, const uint32_t seq_len_kv,
     // A hardcoded 128 here only matches D=128; for D=32/64 the readers would un-swizzle
     // with the wrong pattern and consume out-of-tile shared memory.
     static constexpr uint32_t kSwizzleMode = kHeadDim;
+    DG_STATIC_ASSERT(kHeadDim == 32 or kHeadDim == 64 or kHeadDim == 128,
+                     "kSwizzleMode must stay within the hardware swizzle modes");
     static constexpr uint32_t kSwizzleAlignment = kHeadDim * 8;
     static constexpr uint32_t kSMEMKBytes = kHeadDim;
 
--- a/deep_gemm/include/deep_gemm/impls/sm120_fp8_paged_mqa_logits.cuh
+++ b/deep_gemm/include/deep_gemm/impls/sm120_fp8_paged_mqa_logits.cuh
@@ -53,7 +53,12 @@ void sm120_fp8_paged_mqa_logits(const uint32_t batch_size,
     static constexpr uint32_t kNumMathWarps = kNumMathThreads / 32;
     static constexpr uint32_t kWarpsPerGroup = BLOCK_KV / MMA_M;
     static constexpr uint32_t kNumGroups = kNumMathWarps / kWarpsPerGroup;
-    static constexpr uint32_t kSwizzleMode = 128;
+    // Must match the host-side TMA descriptors (swizzle_mode = head_dim bytes) and the
+    // `tma::copy<..., kHeadDim>` smem layout: the fp8 row is head_dim bytes, so the TMA
+    // writes are swizzled in head_dim-byte atoms (32B/64B/128B for D=32/64/128).
+    static constexpr uint32_t kSwizzleMode = kHeadDim;
+    DG_STATIC_ASSERT(kHeadDim == 32 or kHeadDim == 64 or kHeadDim == 128,
+                     "kSwizzleMode must stay within the hardware swizzle modes");
     static constexpr uint32_t kSMEMKBytes = kHeadDim;
 
     DG_STATIC_ASSERT(kNumTMAThreads == 128, "Expected 128 TMA threads");


aganhui:
Addressed in 5825212 — and thank you for catching this, the original description's "both paged variants are head_dim-128-only" claim was wrong: the `head_dim == 128` assert belongs to the fp4 paged dispatch only (`sm120_fp4_paged_mqa_logits`), while `sm120_fp8_paged_mqa_logits` has no head_dim restriction and its host descriptors pass swizzle_mode = head_dim.

Empirically confirmed on the 5090 D with a targeted fp8 paged repro (small configs; the full paged suite exceeds 32 GB here): accuracy diff vs the test's reference was **0.83 (D=32) / 0.72 (D=64)** while self-consistency *passed* — in the paged variant the misreads land on deterministically-stale shared memory, so the wrong values are bit-stable across runs and only the accuracy comparison exposes them. After deriving kSwizzleMode from kHeadDim: **D=32: 0.0008, D=64: 0.001** (D=128 unchanged).

Also added the suggested `DG_STATIC_ASSERT` pinning head_dim to {32, 64, 128} in both files. Dense regression after the asserts: fp8 D=32 accuracy diff 0.0006.

aganhui:
Cross-reference: leavelet had already identified this bug and prepared an essentially identical one-line fix in #379 (open since 7/15, bot-approved, covering both the dense and paged kernels from the start). We missed that open PR before filing — apologies for the duplication.

Deltas of this branch over #379: the `DG_STATIC_ASSERT` guards and a longer explanatory comment. What #433 adds over both PRs is the evidence/documentation layer: the 189-config failure surface, the symptom forensics (nondeterministic in dense vs *deterministically wrong* in paged — only accuracy comparison exposes the latter), the cross-arch derivation analysis, and the connection to the misdiagnosed reports in #392.

Happy to close this in favor of #379 if maintainers prefer; the derivation is identical and both validate cleanly on our 5090 D. Have posted our validation data to #379 to help it land.

PR #442, opened 2026-09-14
Add torch dispatcher op for tf32_hc_prenorm_gemm

leevan:
## What

Registers `deep_gemm::tf32_hc_prenorm_gemm` as a **torch dispatcher op** (`torch.ops.deep_gemm.tf32_hc_prenorm_gemm`) in addition to the existing pybind binding.

## Why

External C++ (and Python) callers can then invoke this kernel through the PyTorch dispatcher **without including DeepGEMM headers or linking against DeepGEMM** — e.g. a host-side C++ extension can resolve it via `c10::Dispatcher::singleton().findSchemaOrThrow("deep_gemm::tf32_hc_prenorm_gemm", "")`. The schema is compiled into the `_C` extension alongside the pybind bindings, so the op is available as soon as `import deep_gemm` loads `_C`.

## Details

- New header `csrc/apis/torch_ops.hpp`:
  - A thin forwarder to `deep_gemm::hyperconnection::tf32_hc_prenorm_gemm`, converting `std::optional<int64_t>` → `std::optional<int>`.
  - `TORCH_LIBRARY(deep_gemm)` schema: `tf32_hc_prenorm_gemm(Tensor a, Tensor b, Tensor(d!) d, Tensor(s!) sqr_sum, int? num_splits=None) -> ()` — `d` and `sqr_sum` are marked mutable (written in place), `num_splits` defaults to `None` to match the pybind signature.
  - `TORCH_LIBRARY_IMPL(deep_gemm, CUDA)` registering the CUDA implementation.
- `csrc/python_api.cpp`: include the new header so the schema is compiled into `_C`.

### Notes

- The `TORCH_LIBRARY` handle is named `lib` (not `m`) so the `.pyi` stub generator, which scans csrc for pybind-style `m.def(...)`, skips these dispatcher schemas.
- Only a CUDA impl is registered (no Meta/fake kernel); the op is intended for eager execution. Happy to add a Meta kernel if there is interest in tracing through it under `torch.compile`.
- This wraps a single API to establish the pattern; more host APIs can be exposed the same way later if useful.


leevan:
Thanks for the detailed review. Pushed a follow-up commit (9f07225) addressing most of the comments:

- **🟡 multi-TU `TORCH_LIBRARY`**: switched to `TORCH_LIBRARY_FRAGMENT(deep_gemm, lib)`, which is the documented mechanism for splitting schema registration across files. This directly supports exposing more `deep_gemm` ops the same way later without a duplicate-namespace error. Also added a header comment stating the header must be included by exactly one TU.
- **🔵 `num_splits` int64→int narrowing**: added a `TORCH_CHECK` that validates the value is in `[1, INT_MAX]` before the `static_cast<int>`, so an out-of-range value fails loudly instead of silently wrapping.
- **🔵 explicit includes / style**: added `<cstdint>`, `<limits>`, `<optional>` and switched `at::Tensor` → `torch::Tensor` to match the rest of the APIs.
- **🔵 test coverage**: added `test_hc_prenorm_gemm_torch_op` in `tests/test_hyperconnection.py`, which runs the op through the dispatcher (`torch.ops.deep_gemm.tf32_hc_prenorm_gemm`) and compares it against the pybind path for both `num_splits=None` and an explicit split. This guards the schema/forwarder signature against future pybind changes.
- **🔵 Meta/fake kernel**: documented the eager-only limitation in the header (tracing under `torch.compile`/`make_fx` fails with "not implemented for Meta"). Happy to add a Meta kernel in a follow-up if `torch.compile` support for external callers is desired.

- **🔴 GIL / shared JIT state**: this is a real window — the dispatcher entry can be called without the GIL, unlike the pybind binding, so concurrent *first-time* compilation of new shapes depends on DeepJIT's own thread-safety. I've documented the contract in the header (callers should warm up the cache before concurrent use). I'd prefer not to unconditionally `gil_scoped_acquire` here, since (a) it reintroduces a Python dependency that this entry point is specifically meant to avoid, and (b) the right fix may be to make the JIT cache thread-safe rather than serialize on the GIL. Could a maintainer confirm whether DeepJIT's compile cache is expected to be thread-safe? Happy to follow whichever direction you prefer — add a GIL guard here, or track cache-level locking as a separate issue.

Update: verified locally on an SM90 GPU (CUDA 13.2, PyTorch 2.11, Python 3.12, cxx11abi).

- `_C` builds cleanly (only a benign visibility warning on `GilScopedRelease`).
- `import deep_gemm` loads and `torch.ops.deep_gemm.tf32_hc_prenorm_gemm` resolves with the expected schema `deep_gemm::tf32_hc_prenorm_gemm(Tensor a, Tensor b, Tensor(d!) d, Tensor(s!) sqr_sum, int? num_splits=None) -> ()`.
- `tests/test_hyperconnection.py` passes, including the new `test_hc_prenorm_gemm_torch_op`: the dispatcher path matches the pybind path for `num_splits` = None and 16 (calc_diff < 1e-10).

PR #450, opened 2026-09-16
feat(gemm): opt-in labels-contract checker for m-grouped contiguous GEMM

lucifer1004:
## Summary

Add an opt-in host-side contract checker for labels-mode m-grouped contiguous GEMM (`m_grouped_fp8_fp4_gemm_nt_contiguous`, `m_grouped_bf16_gemm_nt_contiguous`, and the `nn` forwarders), turning a silent wrong-results contract violation into a loud, actionable error.

## The contract and the silent failure

The contiguous kernels select B (and SFB) per `BLOCK_M` tile from the label of the tile's **first row** (`grouped_layout[m_block_idx * BLOCK_M]`). The tile heuristics guarantee `BLOCK_M` divides the runtime `mk_alignment_for_contiguous_layout`, so labels whose group boundaries are aligned to the runtime alignment can never put two groups inside one tile.

A caller that builds labels at a **finer** granularity than the current runtime alignment (e.g. labels packed at 64 while the runtime default is 128) puts group boundaries inside a tile. The tile then computes all its rows with the first row's group B — silently corrupting every group after the first misaligned boundary. This was reported downstream as "an empty middle group corrupts later groups" (an empty group merely makes a misaligned boundary much more likely; no empty group is required — `[128, 130, 65]` built at 64 with runtime 128 corrupts the same way).

This is a caller-contract violation, and today it fails silently on every arch (sm90/sm100/sm12x all select B per tile from the first row's label).

## What this PR does

- `csrc/apis/gemm.hpp`: `check_contiguous_labels_contract()` — when `DG_CHECK_CONTIGUOUS_LABELS` is set, validates on the host (one GPU→CPU copy of the labels) that every group's first row starts at a multiple of the runtime mk alignment, and that labels are in range; violations raise a `DGException` naming the group, row, and alignment, with the remedy. Opt-in: zero cost in production.
- `tests/generators.py`: `generate_m_grouped_contiguous` gains an optional `actual_ms_override` (default path byte-identical).
- `tests/test_fp8_fp4.py`, `tests/test_bf16.py`:
  - `test_m_grouped_gemm_contiguous_middle_empty_groups` — empty groups in the middle followed by non-empty groups (e.g. `[100, 0, 130, 65]`), the previously uncovered pattern class, verified contract-respecting;
  - `test_m_grouped_gemm_contiguous_labels_contract_rejection` — misaligned labels raise loudly with the checker on, and pass with the checker off plus matching alignment.

## Validation

- sm_103 (B300): new tests pass; pre-existing `test_m_grouped_gemm_contiguous` (fp8_fp4 + bf16) and `test_m_grouped_gemm_masked` suites pass unchanged (the checker is inert unless enabled).
- sm_120a: rejection path exercised (fires before arch dispatch; main has no sm12x dispatch by design).
- No behavioral change with the env unset; no device-code change.


lucifer1004:
Thanks @ds-review-bot for the careful rounds — all findings addressed in `ba1bcf5` (force-pushed, same scope: checker + tests + docs, no device-code change).

**🟡 warnings — fixed**
- **Resume-after-padding gap (v6, gemm.hpp:199)**: `prev_label` now tracks the previous *row's* label (padding included), so a non-padding run resuming after padding at an unaligned row is rejected. Both patterns from the report are pinned as regression tests: `[0]*128+[-1]*3+[0]*125+...` and `[0]*128+[1]*128+[-1]+[1]*127` at alignment 128 must raise.
- **Positive half swallowing unexpected errors (v6, test_fp8_fp4.py:234)**: the dispatch-failure branch now asserts `get_arch_major() == 12` (the only arch without kernel dispatch on main); supported architectures must run to completion and are numerically verified.
- **alignment=0 division by zero (v4, gemm.hpp:190)**: `DG_HOST_ASSERT(alignment > 0)` added in the checker.
- **README env var (v4/v5)**: `DG_CHECK_CONTIGUOUS_LABELS` documented under *Debug and profiling*.

**🔵 suggestions**
- **Exact per-tile rule (v5, gemm.hpp:194)**: kept the run-based rule deliberately. It is stricter than the kernel's per-tile semantics only for layouts no realistic caller builds (a group split by padding *within* one tile, e.g. `[0]*50+[-1]*10+[0]*68`), and for those it fails loudly with a clear remedy instead of silently accepting — while never accepting a silently-wrong layout. The actual silent-corruption gap (resume after padding) is fixed per v6.
- **Any-negative-is-padding (v4, gemm.hpp:196)**: range check relaxed to `label < num_groups`; comment updated to state the checker matches the device scheduler.
- **Range branch untested (v4)**: out-of-range case added (`label >= num_groups` must raise `out of range`).
- **bf16 entry coverage (v4/v5)**: `test_bf16.py::test_m_grouped_gemm_contiguous_labels_contract_rejection` added and registered.
- **`actual_ms_override` overriding `num_groups` (v5, generators.py:384)**: now `assert num_groups == len(actual_ms_override)`.

**Validation** (fresh JIT cache, `build` artifacts recompiled from this exact tree)
- sm_103 (B300): both rejection tests pass — the fp8/fp4 positive half really dispatches and passes numerics on sm_103; the two resume-after-padding patterns and the out-of-range pattern raise as expected; pre-existing m-grouped contiguous/masked suites (`test_fp8_fp4.py`, `test_bf16.py`) all pass.
- sm_120 (RTX PRO 6000): rejection tests pass; the positive half takes the arch-12 dispatch branch as asserted.

PR #451, opened 2026-09-17
[Bugfix] Reset MegaMoE combine readiness for CUDA graph replay

foreverrookie:
CUDA graph replay can reuse `%gridid`, allowing MegaMoE to accept stale combine-readiness tags and read old or partially updated results.

Reset local peer-readiness flags before the existing pull barrier in both FP8/FP4 and BF16 kernels. The workspace layout and existing barriers are unchanged.

Validation:

- FP8: vLLM adapter tests on 8 B300 ranks, including input staging, unequal/empty source ranks, and changed-input graph replay: 3 tests passed per rank.
- BF16: standalone tests on 2 B300 GPUs, with balanced, unequal, and empty-source cases. Eager execution passed before and after; graph replay failed before the fix, and all 12 changed-input replays per case match eager output exactly after it.

AI assistance was used for the fix and validation.


foreverrookie:
Added the same BF16 reset in 7e924a8. On 2 B300 GPUs, changed-input graph replay failed before the fix in balanced, unequal, and empty-source cases; after the fix, all 12 replays per case match eager output exactly.

PR #458, opened 2026-09-23
Merge main (Public Release 26/09) into nv_dev, with sparse MQA-logits integration fixes

lfr-0531:
## Summary

Bring `nv_dev` onto the DeepJIT-based runtime of the 26/09 public release (`main` 78b6900 = #432 + #441) while keeping every `nv_dev` feature, plus three follow-up fixes found while integrating the sparse MQA-logits kernels into TensorRT-LLM.

### Merge of `main` (26/09) into `nv_dev`

Policy of the merge: `main`'s structure is the base and every `nv_dev` feature is re-applied on top of it.

Now on this branch from `main`:
- DeepJIT submodule replaces `csrc/jit/*` and `fmt` (`jit->compile` / `jit->launch`, C++20, `std::format`); CUDA >= 12.9
- Sparse MQA logits for the DeepSeek-V4.1 hierarchical sparse indexer (`fp8_fp4_sparse_mqa_logits`, `fp8_fp4_paged_sparse_mqa_logits` and their metadata builders)
- MegaGate, Mega mHC, fp8xfp8 MegaMoE weights, `MegaMoESignals` workspace, L2 readiness mask, GEMM alpha / deterministic paths, scheduled MQA metadata
- #441 task-info slot release ordering fence in the MegaMoE scheduler

`nv_dev` features re-applied on the new runtime:
- SM120/SM121: FP8/FP4 GEMM (AB-swap, split-K), BF16 GEMM, bmk/bnk einsum, TF32 HC prenorm GEMM, MQA logits (contiguous + paged), per-arch contiguous M/K alignment, packed UE8M0 SF on SM120
- NVFP4 (fp4xfp4) MegaMoE with BF16 shared experts (`MmaKind::NVFP4` kept next to upstream's `MmaKind::MXF4`; bit-based buffer layout; legacy `l2_full_count` signal for the count protocol)
- SiTU activation for the FP8xFP4 MegaMoE
- SM90 FP8 MegaMoE
- FP16-weights SM100 MQA logits (two-CTA FP16-accumulator kernel)
- SM90 paged MQA logits: block_kv 32, varlen metadata, next_n=4 multicast clusters
- Standalone `smxx_clean_logits` kept for the kernels that do not fuse cleaning (SM120 and FP16-weights paths)
- Barrier timeout policy (Diagnostic / TrapOnly) and the configurable `DG_JIT_BARRIER_TIMEOUT_SECONDS`
- Tests for all of the above merged into the upstream test files

### Follow-up fixes

- `get_sparse_mqa_logits_workspace`: frameworks capture CUDA graphs on a dedicated side stream but replay on the launching stream, so the first sparse-logits call on the capture stream used to trip the pre-capture assertion. While capturing, reuse a workspace already warmed up on the device.
- `get_bf16_mega_gate_config`: return `std::map<std::string, int>` instead of `py::dict` so generated Python stubs are `dict[str, int]` (strict type checkers reject a bare `dict`); the Python-side result is unchanged.
- Paged sparse MQA logits: accept any 16-byte aligned KV page stride. The kernel addresses pages directly and copies them with 16-byte `cp.async` chunks, so page-footer caches that are not padded to 512 bytes (for example 64 x (64 + 4) = 4352-byte pages) can be read in place.

## Validation

GB300 (NGC PyTorch 26.08, CUDA 13.4), standalone test suites: `test_attention` (contiguous MQA logits full, paged sampled cases, sparse MQA logits 161/161 bitwise vs dense), `test_fp8_fp4`, `test_bf16`, `test_einsum`, `test_layout`, `test_hyperconnection`, `test_mega_moe` (4 ranks), `test_nvfp4_mega_moe`, `test_mega_moe_situ`. SM90 and SM120 paths compile but were not run. `test_mega_gate` / `test_mega_mhc` need `tilelang`, which was not available in that environment.

The branch is consumed by TensorRT-LLM's DeepSeek-V4.1 bring-up (CSA2 indexer using the sparse MQA-logits kernels for the candidate-consuming layers), where the DeepGEMM-side changes above were exercised end to end (unit suite, GSM8K, long-context throughput on GB300).

## Note on DeepJIT

`third-party/deep_jit` stays at upstream e5bdee2. Its `utils/exception.hpp` includes `<elfutils/libdwfl.h>` for declarations only (libdw is `dlopen`ed at runtime), so build hosts need the elfutils development headers; making that include optional (`__has_include` with a symbol-only backtrace fallback) is proposed separately to DeepJIT.


lucifer1004:
Hi Fanrong, did you see #447?

lfr-0531:
Duplicated with https://github.com/deepseek-ai/DeepGEMM/pull/447. Closing.

